PR #39, opened 2014-11-21
Recursive record

ericdill:
@arkilic Curious what your thoughts are on this so far...


ericdill:
Ah crap I still need to update the demo scripts

PR #50, opened 2014-12-10
ENH: Added api/analysis.py

ericdill:


ericdill:
@arkilic have a look... temporary find_latest function that I can start beating on

PR #53, opened 2014-12-17
ENH: First pass at converting events -> object

ericdill:
Requested by Claudio and Vivek

### DIFF
--- a/metadataStore/api/analysis.py
+++ b/metadataStore/api/analysis.py
@@ -1,5 +1,5 @@
 __author__ = 'edill'
 
 from ..mdstools import find, find2, find_last
-
+from ..mdstools import as_object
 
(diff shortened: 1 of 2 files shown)


ghost:
I don't mind adding dependencies as long as there is a justification to it. Why would you like to use pandas in mds? I don't think data formatting tools belong in metadataStore itself. We can work on purpose specific utility tools that use metadataStore as a dependency.

ericdill:
Basically the same thing that `as_object` does but do `as_dataframe` (dataframe is the data structure for pandas)

ericdill:
Does `this` refer to the current `as_object` function or the vaporware `as_dataframe` function?

ghost:
It applies to both. We should keep metadataStore itself as simple and refined as we can with simple apis returning python dictionaries. All the other formatting can be distributed to other libraries. This way, once v4 stuff is introduced and metadataStore becomes a service, we won't have to add all these different formatting tools to the service side.

PR #54, opened 2014-12-18
Development

ericdill:


metadataStore/mdstools/__init__.py
```diff
@@ -8,6 +8,16 @@
 from metadataStore.sessionManager.databaseInit import metadataLogger
 from bson.objectid import ObjectId
 from .. import header_version as CURRENT_HEADER_VERSION
+import six
+
+valid_types = (str, int, float, dict, list)
```
ericdill: @arkilic is `tuple` a valid type here too?

PR #56, opened 2015-01-07
DOC: Updated schema docs to prototype frame store

ericdill:
cc @danielballan @tacaswell @stuwilkins @arkilic


frame_hdr_events.txt
```diff
@@ -37,13 +38,25 @@ The return value will be formatted like this:
...
+            type_descriptor: string
+            data_keys: [
+                pv1,
+                ...
+                pvN
+            },
+            external_data_keys: {
+                ext_pv1,
+                ...
+                ext_pvM
```
danielballan: I recall we said that `data_keys` would be all-inclusive, so for illustrative purposes the list of `data_keys` should also include `ext_pv1`, `ext_pv2`, etc. "Internal" data keys can be computed as needed using set logic.

ericdill:
@stuwilkins @dchabot can you guys look at and comment on these proposed changes to the schema?

PR #61, opened 2015-01-26
ENH: Added beamline_id to Event(s) and provided a search routine for broker to search metadataStore Events w/o having to touch any header information

ghost:


metadataStore/commands.py
```diff
@@ -370,6 +370,31 @@ def find_last():
     return Header.objects.order_by('-_id')[0:1][0]
 
 
+def search_events_broker(beamline_id, start_time, end_time):
```
ericdill, line 373: why is the word `broker` in this function name?

ghost, line 373: Because Dan asked for this specifically for the broker. We had a convo this afternoon. It will go sleepy bye tmrw latest!=

tacaswell:
This commits are also part of #62

PR #62, opened 2015-01-26
Add tests

tacaswell:


metadataStore/test/test_commands.py
```diff
@@ -0,0 +1,58 @@
...
+
+def teardown():
+    conn = mongoengine.connection.get_connection('test_db')
+    conn.drop_database(db_name)
+    conn.drop_database(dummy_db_name)
+
+
+def context_decorator(func):
+    def inner(*args, **kwargs):
+        with switch_db(BeamlineConfig, 'test_db'), \
```
ericdill, line 38: use `()` to get rid of the `\`

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+  - source activate testenv
+  - pip install coveralls mongoengine
+  - python setup.py install
+
+
+script:
+  - python run_tests.py
+
+after_success:
+  coveralls
```
ericdill, line 30: hah, you already have the basics of coveralls ready for me

metadataStore/test/test_commands.py
tacaswell, line 38: For reasons that I don't recall right now, you can't do that with context managers.  This use case is the only reason `\` line continuation has not been deprecated.

ericdill, line 38: ah well fair enough

ericdill:
looks fine to me

tacaswell:
Holding off on further tests until after tomorrow morning

PR #65, opened 2015-01-29
Getting closer to the Ineffable One True Spec

danielballan:
Hey @arkilic can we put these commits into master? I know it's still WIP, but it's usable by databroker at this point. Incremental changes from here are expected.


tacaswell:
@arkilic Can you fix up setup.py which is now looking for packages that no longer exist.

ghost:
It is one of the various things that are broken. Working on the refactoring. Give me some time…

metadataStore/commands.py
```diff
@@ -1,93 +1,89 @@
...
+        The date/time as found at the client side when an event is
+        created.
+    beamline_id: str
         Beamline String identifier. Not unique, just an indicator of
         beamline code for multiple beamline systems
-
-    custom : dict, optional
+    owner: str, optional
+        Specifies the unix user credentials of the user creating the entry
+    scan_id : int, optional
```
tacaswell, line 26: scan_id should probably be a string.  I though the current statement was 'anything sortable' and seems safer as a string.

metadataStore/commands.py
```diff
@@ -1,93 +1,89 @@
...
+    if beamline_config is None:
+        begin_run = BeginRunEvent(time=time, scan_id=scan_id, owner=owner,
+                                  time_as_datetime=__convert2datetime(time),
+                                  beamline_id=beamline_id, custom=custom)
+    else:
+        begin_run = BeginRunEvent(time=time, scan_id=scan_id, owner=owner,
+                                  time_as_datetime=__convert2datetime(time),
+                                  beamline_id=beamline_id, custom=custom,
+                                  beamline_config=beamline_config.id)
+    # Had to update to an if/else statement because appending a ReferenceField apparently breaks mongoengine(eye roll)
```
tacaswell: pep8

also, what?

tacaswell:
```
begin_run = BeginRunEvent(time=time, scan_id=scan_id, owner=owner,
                                  time_as_datetime=__convert2datetime(time),
                                  beamline_id=beamline_id, custom=custom,
                                  beamline_config=beamline_config.id if beamline_config else None)
```

should also solve the problem in one call

ghost: see beamline_config. Optional foreignkey. You cannot append a Reference field: begin_run.beamline_config = ObjectId is invalid. So if user creates via beamline_config fall to else

metadataStore/commands.py
ghost, line 26: I need clarification on scan_id. It is optional anyway. It is a simple field change from my perspective.

metadataStore/commands.py
ghost: Are we really that persnickety about our code to the certain point where we care about 1 if statement. That only means code is solid lol. Will change :+1:

metadataStore/commands.py
```diff
@@ -141,78 +125,54 @@ def save_event_descriptor(event_type_id, descriptor_name, data_keys, **kwargs):
...
-    Other Parameters
-    ----------------
-
-    owner: str
-    Specifies the unix user credentials of the user creating the entry
-
-    description: str
-    Text description of specific event
-
+    time : timestamp
```
tacaswell, line 142: Can you put these in the right order and the optional ones as optional?

metadataStore/commands.py
```diff
@@ -238,23 +198,28 @@ def find_header(limit=50, **kwargs):
     unique_id : str
         Hashed unique identifier
 
+    Returns
+    -------
+    br_objects: mongoengine.Document
```
tacaswell: ~~type should be 'list'~~

tacaswell: and missing a ' ' before the ':'

ghost: Actually, I think it should be a QuerySet.

metadataStore/commands.py
```diff
@@ -276,30 +241,34 @@ def find_header(limit=50, **kwargs):
         pass
 
     try:
-        st_time_dict = kwargs.pop('create_time')
-        if not isinstance(st_time_dict, dict):
-            raise ValueError('Wrong format. create_time must include '
+        time_dict = kwargs.pop('time')
+        if not isinstance(time_dict, dict):
```
tacaswell, line 238: Consider doing this as 

``` python
try:
    search_dict['time'] = {'$gte': time_dict['start'],
                                       '$lte': time_dict['end']}
except TypeError:
    raise ValueError('Wrong format. time must be a dict')
except KeyError:
   raise ValueError('time must include start '
                                 'and end keys for range search')
```

metadataStore/commands.py
ghost, line 142: I changed the order but I don't think there is anything optional in event.

metadataStore/commands.py
```diff
@@ -276,30 +241,34 @@ def find_header(limit=50, **kwargs):
...
         pass
 
     if search_dict:
-        header_objects = Header.objects(
-               __raw__=search_dict).order_by('-_id')[:limit]
+        br_objects = BeginRunEvent.objects(__raw__=search_dict).order_by('-_id')[:limit]
     else:
-        header_objects = list()
-    return header_objects
+        br_objects = list()
```
tacaswell, line 254: can you fake this up better so that it always returns the same type of object?  Maybe just raise an exception if the search dict is empty?

metadataStore/commands.py
```diff
@@ -318,23 +287,79 @@ def find_event_descriptor(_id):
     _id: bson.ObjectId
```
tacaswell, line 279: This seems like it is wrong now

metadataStore/commands.py
ghost, line 238: Yup, I have to do this for all time related searches. There are quite a few of those. I think I need to make this a routine and call it instead of typing over and over

metadataStore/commands.py
```diff
@@ -318,23 +287,79 @@ def find_event_descriptor(_id):
...
     connect(db=database, host=host, port=port)
-    event_list = list()
-    for event in Event.objects(header=header.id).order_by('-_id'):
-        event = __replace_event_data_key_dots(event, direction='out')
-        event_list.append(event)
-    return event_list
+    search_dict = dict()
+    try:
+        time_dict = kwargs.pop('time')
+        if not isinstance(time_dict, dict):
```
tacaswell, line 311: Same logic as above with the `try` block.  I think you have enough if staments here that you will never get the `KeyError` raised.

Probably worth pulling this out into a helper function.

metadataStore/odm_templates.py
```diff
@@ -0,0 +1,137 @@
+__author__ = 'arkilic'
```
tacaswell, line 1: :+1: :+1: on this file.

metadataStore/test/test_commands.py
```diff
@@ -99,3 +97,49 @@ def test_dict_key_replace_rt():
...
+
+
+def test_begin_run():
+    bcfg = mdsc.insert_beamline_config({'cfg1': 1})
+    time = ttime.time()
+    beamline_id = 'sample_beamline'
+    yield _begin_run_with_cfg_tester, bcfg, time, beamline_id
+
+
+def _end_run_tester(begin_run, time):
```
tacaswell, line 135: this should have the context decorator

metadataStore/test/test_commands.py
```diff
@@ -99,3 +97,49 @@ def test_dict_key_replace_rt():
...
+    begin_run = mdsc.insert_begin_run(time, beamline_id, beamline_config=beamline_cfg)
+
+    ret = BeginRunEvent.objects.get(id=begin_run.id)
+
+    for k, v in zip(['time', 'beamline_id', 'scan_id', 'beamline_config'],
+                    [time, beamline_id, None, beamline_cfg.to_dbref()]):
+        assert_equal(getattr(ret, k), v)
+
+
+def test_begin_run():
```
tacaswell, line 128: ~~should have context detorater~~ not sure on this one

metadataStore/test/test_commands.py
ghost, line 135: cannot copy paste properly. fixed it

danielballan:
Designing broker/muggler off of a moving target is not working out well, so I'm merging this. Let's distill our various discussions in several separate PRs off of this refactor.

PR #66, opened 2015-02-02
Decorate connection

ericdill:
This PR does a number of things.
1. Do the connection to the database via a decorator so that the connection code is centralized
2. Clean up some of the docs
3. PEP8 line formatting
4. Convert the mds config into a dictionary
5. Fix the cosine test

### DIFF
--- a/metadataStore/commands.py
+++ b/metadataStore/commands.py
@@ -4,10 +4,20 @@ import six
 from metadataStore.odm_templates import (BeginRunEvent, BeamlineConfig,
                                          EndRunEvent, EventDescriptor, Event)
 import datetime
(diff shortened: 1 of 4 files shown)


tacaswell:
:+1: from me.

metadataStore/commands.py
```diff
@@ -109,6 +117,9 @@ def insert_event_descriptor(begin_run_event, data_keys, time, event_type=None):
     data_keys : dict
         Provides information about keys of the data dictionary in
         an event will contain
+    time : timestamp
+        The date/time as found at the client side when an event is
```
danielballan: when an ~~event~~ event descriptor is created

PR #72, opened 2015-02-12
Deployment

tacaswell:
On a clean database the demo works correctly.

### DIFF
--- a/example/synthetic_cosine.py
+++ b/example/synthetic_cosine.py
@@ -1,23 +1,27 @@
-from __builtin__ import type
+from __future__ import print_function
 
 from metadataStore.api.collection import (insert_begin_run, insert_beamline_config,
                                           insert_event, insert_event_descriptor)
 from metadataStore.api.analysis import find_last, find_event, fetch_events
-import random
 import time
-import string
 import numpy as np
-
-
-def id_generator(size=6, chars=string.ascii_uppercase + string.digits):
-    return ''.join(random.choice(chars) for _ in range(size))
 
 
 b_config = insert_beamline_config(config_params={'my_beamline': 'my_value'})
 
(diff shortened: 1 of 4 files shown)


ghost:
Should we update the tests before merging?

tacaswell:
Yes, I am working on them now

tacaswell:
This is going to break the DB at SRX as the events were going in without uids and there is a bunch of synthetic data in there with old-style event_descriptors.

Unless Garth did a whole bunch of scans after @dchabot and I left, we should probably just drop that db and start fresh tomorrow morning.

tacaswell:
@stuwilkins These changes mean that the data from commissioning round 1 will not be accessible to the current version of the tools without a legit database migration effort (probably a day-ish of work), is this a problem?

stuwilkins:
No. We can nuke what we did.

S
On Feb 11, 2015, at 11:18 PM, Thomas A Caswell <notifications@github.com<mailto:notifications@github.com>> wrote:

@stuwilkinshttps://github.com/stuwilkins These changes mean that the data from commissioning round 1 will not be accessible to the current version of the tools without a legit database migration effort (probably a day-ish of work), is this a problem?

stuwilkins:
One other thing which came up speaking to @ericdill, can we make sure that the mongo database is configured correctly. i.e. it can be started and stopped by the sys-v scripts! We found that at CSX this was not the case and when the database started it:

1) Only bound to localhost
2) Saved data in a different directory. 

I think we should ensure that the darabase is also writing to a good place on the system.

Thanks,
Stuart

ericdill:
Have a look at NSLS-II/docs#5. I think that should be a good start for discussions @stuwilkins

PR #81, opened 2015-02-13
ENH: Add Document (formerly BrokerStruct).

danielballan:
This is WIP. Waiting on #77 to proceed.

All documents coming out of MDS will be `metadatastore.Document` objects, stripped of their mongo-specific contents. This also adds a `time_as_datetime` attribute to anything that has a `time` attribute.

PR #82, opened 2015-02-14
BUG: Dump custom field into main document

ericdill:


ericdill:
@danielballan @tacaswell @arkilic This should get merged asap

ericdill:
This is supposed to be against #81 ...

PR #83, opened 2015-02-15
API : remove time_as_datetime from Event

### LINKED ISSUE #78: remove time_as_datetime from Event
We do not want to support this long term and removing it later will require data migration. This needs to be done as soon as possible before we have lots of data to migrate.

tacaswell:
closes #78


ericdill:
Looks good

PR #86, opened 2015-02-15
API: Remove nested api

ericdill:
IMO, having `api` as the only sub-package does not pass the sniff test. Also, only return sanitized mongo objects from all find functions.

This will break ophyd and ~~metadatastore~~ dataportal, but those are very easy to fix. I will deal with the changes to both projects if everyone agrees that this PR is a good move.

thoughts @danielballan @tacaswell @arkilic

### DIFF
--- a/example/synthetic_cosine.py
+++ b/example/synthetic_cosine.py
@@ -16,7 +16,8 @@ data_keys = {'linear_motor': {'source': 'PV:pv1',
                                  'shape': None,
                                  'dtype': 'number'},
              'Tsam': {'source': 'PV:pv3',
-                      'dtype': 'number'}
+                      'dtype': 'number',
+                      'shape': None}
              }
 
 try:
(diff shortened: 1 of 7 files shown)


tacaswell:
:+1: from me, need to fix the setup.py to not try to install a package that no longer exists (that is the source of travis failure)

ericdill:
good catch, thanks

ghost:
:+1:  :+1: from me!

ericdill:
Do we want the insertions to return `Document` objects too?

ericdill:
I think I vote for yes

tacaswell:
So long as it doesn't break the ability to chain them.

ericdill:
How do you mean "chain them"

tacaswell:
``` python
blc = mds.insert_beamline_config(...)
bre = mds.insert_begin_run_event(blc, ...)
```

danielballan:
Agreed all around

ericdill:
Hmm... making the insertions return sanitized documents is a pretty large pain and, tbh, I'm not entirely sure how to do it.  Let's leave well enough alone for now

ghost:
It really is painful! I think I've an idea but have to search the Internet if its the best way to do it.

ericdill:
@arkilic not sure it's high priority right now

ghost:
No worries. getting rid off api package isn't high priority in the first place :)

tacaswell:
Can you also update all of the docstrings?  Should be able to do it via find-replace?

ericdill:
Which docstrings?

PR #89, opened 2015-02-17
ENH: Implemented find_run_stop

ericdill:
- Also renamed RunEnd -> RunStop


ghost:
:+1: from me!

danielballan:
Great. For the record, the reasoning here is that "start and stop" belong together and "beginning and end" belong together. The current behavior (before this PR) mixed them with "start" and "end." See for example Python slice objects, which use start and stop. This PR matches that convention.

PR #90, opened 2015-02-17
Mds rename

ericdill:
Passes tests locally on python ~~3~~ 2


ericdill:
Approved by @dchabot  @tacaswell @danielballan,

 @arkilic ?

danielballan:
Per conversation with @ericdill, we can do this now while we are repacking everything anyway. (We will also do fileStore -> filestore.) If we want to actually change the name (datastore or some better idea) we can worry about that later.

PR #97, opened 2015-02-20
MNT : add explicit db alias

tacaswell:
So that we can import multiple things that use mongoengine and
independently control which host/database/post each set of
odm templates looks at.

If the alias is not explicitly specified it falls back to using the
default one which is the source of the conflicts as the connection
related state is held as module-level global state in mongoengine
it will be shared by all odm_templates in the same python process.

This change should be completely transparent to all existing code.

### DIFF
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -3,14 +3,14 @@ from __future__ import (absolute_import, division, print_function,
 import six
 from functools import wraps
 from .odm_templates import (RunStart, BeamlineConfig, RunStop,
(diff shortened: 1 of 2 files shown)


danielballan:
While you are there, can you rename `mds_config` to `connection_config` to match filestore?

tacaswell:
:-1: on than renaming.  Does not gain much and just adds (yet) more churn.

danielballan:
OK OK. Churn bad.

PR #99, opened 2015-02-26
FIX: pop raises a KeyError; it does not default to None.

danielballan:
This fixes a bug caught by a test in dataportal. Needed before I can merge NSLS-II/dataportal#80.


ericdill:
ah, that's my bad. I think i screwed that up

danielballan:
No problem. Thanks.

PR #100, opened 2015-02-27
API: Add mds_setup, mds_teardown testing utils.

danielballan:
Just like NSLS-II/filestore#32 but for metadatastore.

Also, this copies the `db_connect`, `db_disconnect`, `_ensure_connection` machinery from FS and standardizes the names.

And finally, it looks like some tests were broken but not being correctly run. This fixes that.


metadatastore/conf.py
```diff
@@ -7,14 +7,14 @@
                         'connection.yml')
 if os.path.isfile(filename):
     with open(filename) as f:
-        mds_config = yaml.load(f)
+        connection_config = yaml.load(f)
```
tacaswell, line 10: And the name thrashing continues

PR #105, opened 2015-03-02
WIP: Do not merge until everybody agrees to changes! 

ghost:


ghost:
Not all the changes we discussed are implemented yet. Trying to take vacation and do homework, will do the rest as soon as I can. Please give me feedback on whatever is implemented. Passed tests on two of my Macs so far...

metadatastore/commands.py
```diff
@@ -548,25 +475,29 @@ def find(data=True, limit=50, **kwargs):
...
+    try:
+        run_start_id = kwargs.pop('run_start').id
+        event_descriptors = EventDescriptor.objects(__raw__=run_start_id)
+    except KeyError:
+        pass
+    if event_descriptors:
+        for e_desc in event_descriptors.order_by('-time'):
+            query_dict = dict()
+            query_dict['descriptor_id'] = event_descriptor.id
+            query_dict.update(kwargs)
```
tacaswell: This is probably not the order you want this precedence to go.

ghost: Hmmm, I wasn't sure how we wanted to handle find_event(run_start=run_start_obj). What do you have in mind?

tacaswell: There is a degeneracy, if you pass in both `run_start` and `descriptor_id`.  As written, the kwarg will always win and you will get back a bunch of copies of the events rather than all of the events.  Can't this all be done in one mongo call?  http://docs.mongodb.org/manual/reference/operator/query/in/#op._S_in

ghost: Yup, we can do it in a single db trip. In a similar fashion, we can make all similar iterative searches to utilize $in functionality.

PR #107, opened 2015-03-05
ENH: Find events by the run start id

ericdill:
- Related to NSLS-II/dataportal#98
- I'm not at all tied to the name, so thrash away

### DIFF
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -486,8 +486,6 @@ def find_event(run_start):
     ---------
     run_start: mongoengine.Document
         RunStart object that events possibly fall under
-    limit: int
-        Number of headers to be returned
 
     Returns
     -------
@@ -495,7 +493,25 @@ def find_event(run_start):
         Set of events encapsulated within a RunStart's scope.
         metadatastore.document.Document
     """
-    descriptors = EventDescriptor.objects(run_start=run_start.id)
+    return find_event_by_runstart_id(run_start.id)
+
+
+@_ensure_connection
+def find_event_by_runstart_id(run_start_id):
+    """Returns a set of events given a RunStart object
+
+    Parameters
+    ---------
+    run_start_id: mongoengine.ObjectID
+        RunStart object that events possibly fall under
+
+    Returns
+    -------
+    events: list
+        Set of events encapsulated within a RunStart's scope.
+        metadatastore.document.Document
+    """
+    descriptors = EventDescriptor.objects(run_start=run_start_id)
     descriptors = descriptors.order_by('-_id')
     events = [find_event_given_descriptor(descriptor)
               for descriptor in descriptors]
--- a/metadatastore/api.py
+++ b/metadatastore/api.py
@@ -1,7 +1,8 @@
 # Data retrieval
 from .commands import (find, find_event, find_event_descriptor,
                        find_beamline_config, find_run_start, find_last,
-                       find_event_given_descriptor, fetch_events, find_run_stop)
+                       find_event_given_descriptor, fetch_events, find_run_stop,
+                       find_event_by_runstart_id)
 # Data insertion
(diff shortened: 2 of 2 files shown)


danielballan:
Thrashing away: `event` -> `events` because it return a list. Otherwise great.

ericdill:
I'm matching the existing API: https://github.com/NSLS-II/metadatastore/pull/107/files#diff-86f2983c849d2ca24f0c14643a623d33R482

metadatastore/commands.py
```diff
@@ -495,7 +493,25 @@ def find_event(run_start):
...
+    return find_event_by_runstart_id(run_start.id)
+
+
+@_ensure_connection
+def find_event_by_runstart_id(run_start_id):
+    """Returns a set of events given a RunStart object
+
+    Parameters
+    ---------
+    run_start_id: mongoengine.ObjectID
```
tacaswell, line 505: This probably should be a string, not an `ObjectID`

ericdill, line 505: @arkilic can confirm this, but I believe _either_ will work here?

danielballan:
That's true, but the existing API is wrong. :- D

ghost:
My pr does this. Take a look at PR #105. Find routines are modified per our agreement

metadatastore/commands.py
ghost, line 505: This works using the method Tom and I have agreed upon earlier:

https://github.com/arkilic/metadatastore/commit/b70253e53318369a3702e600af197be16056be07

danielballan:
@arkilic, I don't think that's true. Your PR does not give us a way to look up Events given a run start ID; it only lets us look up a Events with a RunStart Document that has an `id` attribute. (It looks like that `id` lookup happens in `find_event_descriptor`, which is called by `find_event`.)

If I'm misunderstanding, can you spell out how we should be looking up Events given a run start ID string? Forgive my tired brain.

ghost:
That's the unified way we have agreed upon. We throw run_start objects at all find routines. It was part of the spec. If you are going to use the _id of run start, it should be a bson.ObjectId. This shouldn't be a separate routine, it should be part of find_event. Also, you can always access run_start objects for a given event by eventobj.descriptor.run_start. You don't need have to have a separate routine that does it, mongoengine does it for you using the DbRef field

On Mar 5, 2015, at 7:26 PM, Dan Allan <notifications@github.com<mailto:notifications@github.com>> wrote:

@arkilichttps://github.com/arkilic, I don't think that's true. Your PR does not give us a way to look up Events given a run start ID; I only lets us look up a Events with a RunStart Document that has an id attribute. (It looks like that id lookup happens in find_event_descriptor, which is called by find_event.)

If I'm misunderstanding, can you spell out how we should be looking up Events given a run start ID string? Forgive my tired brain.

danielballan:
@arkilic OK, thanks for reviewing that. The issue we are trying to solve is that DataBroker returns Headers, which are not exactly RunStart objects because they have _two_ IDs, inherited from RunStart and RunEnd, and thus no single `id` attribute. We were thinking we should add a `find_events_from_header` or `find_event_from_run_start_id` to make this work. The alternative -- mutating a Header to look enough like a RunStart to satisfy `find_event` -- was too ugly. What do you think we should do?

ghost:
It's a simple addition to the find_event I put together. I have already done it in an optimized way but with the run_start object as you said. Would you guys be happy if find_event accepted a run_start_id field as string that returned you all the corresponding events for that run_header? I forgot about __as_document goody we added earlier, my bad.

danielballan:
Yes, I think that would do the job.

ghost:
I'll have dinner and add it to my PR.

danielballan:
There's no fire here. It can wait for tomorrow morning.

tacaswell:
Either in #108 or punted to dataportal

PR #108, opened 2015-03-07
Revisions

danielballan:
This is a big one. Bear with me while I try to give a readable overview.

This PR contains:
- commits in @arkilic's master branch and his PR #105, adding functionality agreed on
- commits that @ericdill and I "pair-coded" in the conference room today
- more commits from me, adding tests and fixes of bugs revealed by those tests
### Changes that impact actual functionality
- All `find_*` functions now accept a time range (`start_time`, `end_time`), not a specific `time` to match.
- Every `find_*` function that accepts a document (e.g., `find_run_stop(run_start=run_start)`) also accepts a mongo ID (e.g., `find_run_stop(run_start_id=run_start_id)`) and parses it whether it be a string or a bson `ObjectId`. Precedence is specified in the docstrings.
- All `find_*` functions uniformly return lists.
- `find_last` and `find_run_start` do not dereference their EventDescriptors, as this is being done in the broker by `_build_headers`.
- `find_*` functions that accept a `limit` default to `limit=None`. A constant default (`limit=50`) is now applied at the broker level, and two layers of defaults can be confusing. ("MDS just does what you tell it to do; broker does more to protect the user.")
- Looking at the implementation of `find_event`, the function had become to unwieldy: potentially, one might start asking it to filter Events based on any attribute in the entire schema. To be clear, @arkilic did exactly what we agreed on. Upon seeing it concretely in #105, @ericdill and I changed our minds about what we want. Or new philosophy is this: `find_*` functions can search on documents that they directly reference and attributes that they have. They do not traverse more than one level of referencing. For example, if you want to find Events based on a RunStart, you have to do it in two steps -- using both `find_event_descriptor(run_start=run_start)` and `find_event(descriptor=descriptor)`. You cannot `find_event(run_start=run_start`) because Event does not directly reference RunStart. You cannot `find_event(owner=...)` because Event does not have an owner attribute. A cousin PR to databroker demonstrates this new usage of the `find_*` functions. I think this API is nicely usable and clean. And this plan prohibits the `find_*` functions from growing into a burgeoning wishlist of if/elif/elif....
### Docstring revisions
- Made docstrings correctly reflect current usage
- Explained common Parameters (start_time, uid, etc.) consistently throughout commands.py
### Testing and Debugging
- Every Parameter in the docstrings is exercised by a test. (Exception: `find_event`, which has never had any coverage. @arkilic or I can tackle this next week.)
- Every `insert_*` function issues a DEBUG-level log message include useful ID string(s). According to my understanding, the performance cost is minimal when DEBUG-level logging is turned off. If the cost becomes an issue, we can remove it later, but the logging was very useful to me today.
### Questions
- Now that every `find_*` function returns a list, it seems like a good time to pluaralize all their names. Are you OK with this @arkilic? This would not affect any `insert_*` functions, so it would only touch MDS and broker.
- @arkilic, we documented `_id` as a Parameter, and it works. But in the tests I see that you passed `id` and it seems to work. Are `_id` and `id` synonyms in a mongo query? If so, I will change every `_id` docstring to `id`.


metadatastore/test/test_commands.py
```diff
@@ -60,53 +68,87 @@ def _ev_desc_tester(run_start, data_keys, time):
             assert_equal(getattr(ret.data_keys[k], ik),
                          data_keys[k][ik])
 
+    # Exercise documented Parameters.
+    mdsc.find_event_descriptor(run_start=run_start)
+    mdsc.find_event_descriptor(run_start_id=run_start.id)
+    mdsc.find_event_descriptor(run_start_id=str(run_start.id))
+    mdsc.find_event_descriptor(run_start_id=str(run_start.id))
```
tacaswell: duplicate line

ghost:
> Now that every find_\* function returns a list, it seems like a good time to pluralize all their names. Are  you OK with this @arkilic? This would not affect any insert_\* functions, so it would only touch MDS and  broker: 

From a service point of view, having more routines only slows your service down and complicates service implementation. You would have to figure out which request gets routed where. The more decisions you have to make, the slower your service and the more complicated your service implementation get. Dumping a json load into a find_\* is much simpler than figuring out what the contents of that json load is and routing the request to a routine. When you have multiple clients trying are sending bunch of requests to your service, this gets messy.

> @arkilic, we documented _id as a Parameter, and it works. But in the tests I see that you passed id and it seems to work. Are _id and id synonyms in a mongo query? If so, I will change every _id docstring to id.

No, they are not the same thing. id is how you access to _id field of a mongoengine document. When you submit the raw query to mongo daemon, it expects an _id field. 

> db.event.find({"_id" : ObjectId("54f9ac429ef3a75e45d4b430")})
> { "_id" : ObjectId("54f9ac429ef3a75e45d4b430"), "descriptor_id" : ObjectId("54f9ac429ef3a75e45d4b41c"), "uid" : "eab9bcb9-3936-4236-9064-9131c10a6d4c", "seq_num" : 19, "data" : { "linear_motor" : [ 0.19019019019019018, 1425648706.129566 ], "scalar_detector" : [ 0.9807289108621545, 1425648706.129575 ], "Tsam" : [ 5.19019019019019, 1425648706.129568 ] }, "time" : 1425648706.129577 }
> 
> db.event.find({"_id" : ObjectId("54f9ac429ef3a75e45d4b430")}).count()
> 1
> db.event.find({"id" : ObjectId("54f9ac429ef3a75e45d4b430")}).count()
> 0

This is all good stuff but before we spend more effort to make this API perfect and Mongo-centered, we should really spend effort in stress testing. My experience in sharding mongo is not good at all and I see other people have similar experiences scaling mongo... I have some benchmark code written, let's find the boundaries of what we have before we spend a lot of effort in making something that we might not use perfect.

danielballan:
I don't understand how your first response about avoiding "more routines" relates to my question about changing the naming of our existing routines.

As for avoiding perfecting the mongo implementation, point taken. I consider this PR basic, essential cleanup, but we can avoid optimizing the performance until you stress test mongo.

ghost:
Ahhh, I thought what you meant by pluralize referred to our earlier convo about writing routines in the nature of find_event_given_blah(). Having said that, I am OK with the name change. 

That is my first task once I am back on Thursday. Could you take a look at the below repo @danielballan  @tacaswell ?
https://github.com/arkilic/mdsbenchmark

Far from complete but it is a start...

Issue #102, opened 2015-02-27
Someday: Return generator that makes Event Documents from mongo Cursor


danielballan:
When there are huge numbers of Events:

```
(Document(e) for e in Cursor)
```

tacaswell:
How did I close this?

danielballan:
No idea.

ghost:
@tacaswell had no @tacaswell to oversee his actions. Open loop lol

danielballan:
When @tacaswell errs, Zeus himself strikes down the offending commit. :zap:

ghost:
This conversation is why people tell this sort of stuff about us: http://rs2img.memecdn.com/engineers-personality_o_198312.webp

PR #110, opened 2015-03-09
ENH: find_* functions return generators, have no limit.

danielballan:
This refactors each `find_*` function to return a generator instead of a list. Queries that return a huge number of results are no problem unless you actually try to look at (loop through) those results, so there is no need for `limit`. This will make all queries feel snappier by deferring the Python object creation.

This is based on some discussion with @dchabot, @tacaswell, and @ericdill. There is no urgency -- we will not deploy this tomorrow -- so it can wait for the triumphant return of @arkilic.

### DIFF
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -2,6 +2,7 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 import six
 from functools import wraps
+from itertools import count
 from .odm_templates import (RunStart, BeamlineConfig, RunStop,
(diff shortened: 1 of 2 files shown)


danielballan:
TODO: Benchmark.

ericdill:
Note that this closes #102

danielballan:
@arkilic Would you ceremonially merge this and the cousin PR in databroker?

ghost:
:tada:

danielballan:
Hahahah

Issue #111, opened 2015-03-10
The insert_run_start api doesn't match the RunStart Document


ericdill:
The insert_run_start function would suggest that time and beamline_id are the only required RunStart fields:

```
def insert_run_start(time, beamline_id, beamline_config=None, owner=None,
                     scan_id=None, custom=None, uid=None):
```

However, the fields in RunStart are this:

```
beamline_config, required
owner, required
scan_id, required
time, required
uid, required

beamline_id, optional
group, optional
project, optional
sample, optional
```
- Owner is an optional kwarg to insert_run_start because we can programmatically get the logged in user
- beamline_config _could_ be guessed by doing something like grabbing the last beamline_config that mongo knows about, but that feels really dangerous
- scan_id _could_ be guessed by grabbing the last run_start and incrementing its scan_id by 1

However, these do not feel like terribly good ideas, as it should not be the job of mds to guess what the user wanted to do. That's what ophyd is for!

In the mean time, `scan_id` and `beamline_config` should not be kwargs. However, in the interest of @tacaswell 's sanity this change should wait until after tomorrow's deployment.

ericdill:
Proof that this is a bug: http://nbviewer.ipython.org/gist/anonymous/46cfb2e65810b7c9036b

danielballan:
Closed by #118

cowanml:
In a user exposed api seems like we should _only_ be getting the owner programatically?

danielballan:
Do you mean that we should not allow the user to specify owner -- that we should require it to match the username of the logged in user?

There's a confounding issue here: most (all?) beamlines are set up with a generic beamline user (e.g., 'xf23id1') that just stays logged in no matter who is taking data. If we want users and beamline scientists to fastidiously log in and log out, I expect we'll have to convince them to change their habits. It's possible, but it's not going to happen in the short term. For now, 'owner' is a convenient user-defined tag, not guaranteed to be correct or authenticated.

dchabot:
> On Mar 24, 2015, at 10:03 AM, Dan Allan notifications@github.com wrote:
> 
> Do you mean that we should not allow the user to specify owner -- that we should require it to match the username of the logged in user?
> 
[...]

Issue #85, opened 2015-02-15
Sort out which `insert_*` functions should take custom dicts


tacaswell:
To fully exploit all of the DynamicDocuments we have, we need a way to provide a way to shove random extra data in from the command api.

See #84 for a template.

cc @ericdill @danielballan @arkilic

ericdill:
Shouldn't **all** DynamicDocuments take a `custom` argument?

tacaswell:
Yes.  This is a valid way to sort this out.

The title is just my normal extra-hedgey way of phrasing things when tired/stressed/in hurry.

danielballan:
Closed by #118

PR #118, opened 2015-03-23
FIX: insert_* commands should accept custom fields

danielballan:
- `insert_run_start` already implemented `custom`.
- `insert_run_stop` and `insert_event_descriptor` did not, but they should.
- `insert_beamline_config` did not, but it already accepts an arbitrary dictionary, so there is no need for `custom`.

Also, the `uid` and `time` arguments were not documented everywhere. This fixes that.

### DIFF
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -203,7 +203,7 @@ def insert_beamline_config(config_params, time, uid=None):
 
     Parameters
     ----------
-    config_params : dict, optional
+    config_params : dict
         Name/value pairs that indicate beamline configuration
         parameters during capturing of data
     time : timestamp
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -166,7 +166,7 @@ def insert_run_stop(run_start, time, uid=None, exit_status='success',
     ----------
     run_start : metadatastore.odm_temples.RunStart
         Foreign key to corresponding RunStart
-    time : timestamp
+    time : float
         The date/time as found at the client side when an event is
         created.
     uid : str, optional
@@ -208,7 +208,7 @@ def insert_beamline_config(config_params, time, uid=None):
     config_params : dict
         Name/value pairs that indicate beamline configuration
         parameters during capturing of data
-    time : timestamp
+    time : float
         The date/time as found at the client side when the
         beamline configuration is created.
     uid : str, optional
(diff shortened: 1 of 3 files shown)


danielballan:
Closes #85

metadatastore/commands.py
```diff
@@ -197,6 +206,11 @@ def insert_beamline_config(config_params, time, uid=None):
     config_params : dict, optional
```
ericdill: According to the function signature, `config_params` is not optional

danielballan:
Added more commits to close #111 while I'm at it.

danielballan:
OK, I'm done. This is ready to review.

metadatastore/commands.py
```diff
@@ -167,8 +169,15 @@ def insert_run_stop(run_start, time, exit_status='success',
     time : timestamp
```
ericdill: `timestamp` is inconsistent with line 117

ghost:
Looks good to me!

metadatastore/commands.py
```diff
@@ -155,8 +157,8 @@ def insert_run_start(time, beamline_id, beamline_config=None, owner=None,
 
 
 @_ensure_connection
-def insert_run_stop(run_start, time, exit_status='success',
-                    reason=None, uid=None):
+def insert_run_stop(run_start, time, uid=None, exit_status='success',
```
ericdill, line 160: I thought we were using the pattern "optional args are always `None`". Why is this different here?

metadatastore/commands.py
```diff
@@ -194,9 +205,14 @@ def insert_beamline_config(config_params, time, uid=None):
 
     Parameters
     ----------
-    config_params : dict, optional
+    config_params : dict
```
ericdill, line 208: config_params is optional according to the `BeamlineConfig` document in mds

ericdill, line 208: Granted, that's a little dumb

metadatastore/commands.py
danielballan, line 160: I don't agree that optional args should always be None. It's a good pattern, but it's not the only good pattern. One advantage of _not_ doing it that way is that the user can see from the signature what the default is. I think that makes sense here.

ericdill:
looks good!

Issue #112, opened 2015-03-11
BUG: Fully implement and document 'project' and 'group' in commands


danielballan:
`insert_run_start` does not accept or document the `project` or `group` fields. `find_run_start` does not document `group` as an accepted argument.

danielballan:
Closed by #119

PR #119, opened 2015-03-23
Expose 'group' and 'project' to insert_run_start

### LINKED ISSUE #91: Project field in insert_run_start
The `project` field in the RunStart object needs to be exposed through the `insert_run_start` function

danielballan:
Closes #91 and #112. Merge after #118 -- this builds on commits in #118.

### DIFF
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -108,8 +108,8 @@ def format_events(event_dict):
 
 @_ensure_connection
 def insert_run_start(time, scan_id, beamline_id, beamline_config, uid=None,
-                     owner=None, custom=None):
-    """ Provide a head for a sequence of events. Entry point for an
+                     owner=None, group=None, project=None, custom=None):
+    """Provide a head for a sequence of events. Entry point for an
     experiment's run.
 
     Parameters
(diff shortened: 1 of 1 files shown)


metadatastore/commands.py
```diff
@@ -117,18 +117,24 @@ def insert_run_start(time, beamline_id, beamline_config=None, owner=None,
...
         The date/time as found at the client side when an event is
         created.
-    beamline_id: str
+    scan_id : int
+        Unique scan identifier visible to the user and data analysis
+    beamline_id : str
         Beamline String identifier. Not unique, just an indicator of
         beamline code for multiple beamline systems
-    beamline_config: metadatastore.odm_temples.BeamlineConfig, optional
+    beamline_config : metadatastore.odm_temples.BeamlineConfig
```
ericdill, line 125: ~~maybe make this object `metadatastore.api.BeamlineConfig`~~

nevermind

tacaswell:
Can you push a dummy commit to this and then forcepush it out of existence?  I think gh needs to be kicked to re do the diffs.

tacaswell:
and does this or #118 have companion PRs into ophyd?

danielballan:
@tacaswell Github has been kicked. This does not have a companion PR in ophyd. But I believe @dchabot has some commits making it possible to pass arguments through `*scan` to `insert_run_start`, making all of these fields exposed, if not documented.

tacaswell:
I was more worried about the fact that the order of the args changed.

danielballan:
This is why we need Travis set up for ophyd. I think we were careful to use
kwargs but we should check (tomorrow) before we merge.

danielballan:
Confirmed: the only call to `insert_run_start` in ophyd passes everything by keyword argument. This change is safe.

danielballan:
And likewise with the other `insert_*` calls, whose `uid` kwarg has been reordered by this PR.

PR #120, opened 2015-03-24
API: Reorganize Event.data into data, timestamps.

danielballan:
See NSLS-II/databroker#129 for explanation. Also, there will be a cousin PR in dataporal so that this change does not break broker and muxer. (Replay is unaffected.)


danielballan:
@arkilic How does this look to you?

danielballan:
@arkilic Do a hear a "merge"? :- D

ghost:
I talked to @dchabot  about this. We should talk about if we want to enforce a specific format for incoming data or do it here. Don't merge yet :hand:  :)

PR #121, opened 2015-03-30
BLD: Use container for faster Travis builds.

danielballan:


danielballan:
Travis master broke become mongoengine released 0.9. This fixes it. Merging own, as this only touches `.travis.yml`.

PR #122, opened 2015-03-30
Human times

danielballan:
You can:
- `find_*(start_time='2015')`
- `find_*(start_time='2015-03')`
- `find_*(start_time='2015-03-30 18:00:00')`
- `find_*(start_time=datetime.datetime(...))`

and likewise, of course, for `end_time`.

MDS now expects a `timezone` field in its config file, such as `timezone: 'US/Eastern'`.

Tests are included.


metadatastore/commands.py
```diff
@@ -368,13 +369,23 @@ def _format_time(search_dict):
...
     if time_dict:
         search_dict['time'] = time_dict
 
 
+def _normalize_human_friendly_time(val):
+    "Parse '2015', '2015-03', '2015-03-30', and '2015-03-30 18:00:00'."
+    tz = conf.connection_config['timezone']  # e.g., 'US/Eastern'
+    if isinstance(val, six.string_types) or isinstance(val, datetime.datetime):
+        return pd.to_datetime(val).tz_localize(tz).value/1e9
+        return pd.to_datetime(val).tz_localize(tz).value/1e9
```
tacaswell: probably only need to return once.

danielballan:
@arkilic I'd like to get this merged in time for tomorrow's tag and deployment to all the beamlines.

metadatastore/commands.py
danielballan: That's "probably only need to return once, _doctor_."

tacaswell:
Can you also update the conda-recipe for the new pandas dep

ericdill:
It seems worth adding a note somewhere regarding valid options for this new `timezone` field.  Something like 

``` python
# list the valid timezones for metadatastore config
import pytz
print(pytz.all_timezones)
```

Otherwise :+1:

tacaswell:
Timezone names are pretty standard, not sure it is worth documenting

ghost:
Can we please not add any pandas dependency to metadatastore? Human friendly time zone stuff belongs to broker, not here.

danielballan:
I disagree, but I can see your point of view. I considered putting this in Broker instead but -- my pandas fandom aside -- pandas is part of the core scientific Python stack these days. Why not depend on it? Human-friend times make forensics like the recent XPD debacle much easier.

@ericdill @tacaswell Agreed that the names are standard and not Python-specific, but I think Eric's right that we might as well point people to a list of them.

ghost:
Two reason: 

1: we might change the backend to some language other than Python. Less language specific data structure makes it easier to change things around. There are others in the controls community willing to deploy this in a larger scale and Python will not be the way they go. If they do something better and faster using our language independent architecture, I would like us to be able to easily adopt it.
2: this is a database, not a scientific tool. Jsonifying complex objects are harder and makes rest interface much slower. I have implemented a RESTful service using our current api and it's very fast and literally 100 lines of code because the objects we put on the wire are simple as we designed(good job everybody!). RESTful services sort of solve the issue of how we will get user their data. 

Forensics are their fault, we can't make these backends human fault free. There're too many ways to add human friendly times and I don't think the reward is worth adding a new dependency like pandas.

tacaswell:
The point of pulling in pandas here is _just_ to get it's timestring <->
timestamp parsing.  If we have a lighter library which will do this then we
should rip pandas out, but for now this is the most expedient option.

ericdill:
I'm actually with @arkilic on this one.  I don't see the value in letting mds understand human friendly times.  That really sounds like the job for databroker which converts the human friendly argument to a timestamp and passes it to mds

ghost:
My new communication technique works, I now make sense as of release 28.1 :tada: 
So we need to address two issues: 
Do we want metadatastore to understand human friendly time? If so, can we use pure pytz or some expedient library?

danielballan:
With @arkilic's blessing in an in-person conversation, we are merging this to get the functionality and finding a better package later.

PR #123, opened 2015-03-31
REF: Load configs from env, /etc, and ~/config in precedence.

danielballan:
This PR also refactors the config-loading logic into a function that could be reused by filestore if we can decide on a place to put it. The simple implementation of this function is not compatible with a _nested_ YAML file, but we'll solve that problem when we come to it.

### DIFF
--- a/metadatastore/conf.py
+++ b/metadatastore/conf.py
@@ -3,18 +3,31 @@ import yaml
 import logging
 
 logger = logging.getLogger(__name__)
-filename = os.path.join(os.path.expanduser('~'), '.config', 'metadatastore',
-                        'connection.yml')
-if os.path.isfile(filename):
-    with open(filename) as f:
-        connection_config = yaml.load(f)
-    logger.debug("Using db connection specified in config file. \n%r",
-                 connection_config)
-else:
-    connection_config = {
-        'database': 'test',
-        'host': "localhost",
-        'port': 27017,
-        }
-    logger.debug("Using default db connection. \n%r",
-                 connection_config)
+
+def load_configuration(name, prefix, fields):
+    filenames = [os.path.join('etc', name + '.yml'),
+                 os.path.join(os.path.expanduser('~'), '.config',
+                              name, 'connection.yml'),
+                ]
+    if 'CONDA_ETC_' in os.environ:
+        filenames.insert(0, os.path.join(os.environ['CONDA_ETC_'], name + '.yml'))
+
+    config = {}
+    for filename in filenames:
+        if os.path.isfile(filename):
+            with open(filename) as f:
+                config.update(yaml.load(f))
+            logger.debug("Using db connection specified in config file. \n%r",
+                         config)
+
+    for field in fields:
+        var_name = prefix + '_' + field.upper().replace(' ', '_')
+        try:
+            config[field] = os.environ.get(var_name, config[field])
+        except KeyError:
+            raise KeyError("The configuration field {0} was not found in any "
+                           "file or environmental variable.")
+    return config
+
+connection_config = load_configuration('metadatastore', 'MDS',
+    ['host', 'database', 'port', 'timezone'])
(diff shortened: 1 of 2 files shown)


danielballan:
(Once this is reviewed, I'll put a cousin PR into filestore. Goodnight, Neverland!)

metadatastore/conf.py
```diff
@@ -3,18 +3,29 @@
...
-    connection_config = {
-        'database': 'test',
-        'host': "localhost",
-        'port': 27017,
-        }
-    logger.debug("Using default db connection. \n%r",
-                 connection_config)
+
+def load_configuration(name, prefix, fields):
+    filenames = [os.path.join('etc', name + '.yml'),
```
tacaswell, line 8: This needs to also grab `CONDA_ETC_`

metadatastore/conf.py
```diff
@@ -3,18 +3,29 @@
...
+                 os.path.join(os.path.expanduser('~'), '.config',
```
ericdill, line 9: Did you intentionally drop 'metadatastore' from this path?

metadatastore/conf.py
```diff
@@ -3,18 +3,29 @@
...
-    with open(filename) as f:
-        connection_config = yaml.load(f)
-    logger.debug("Using db connection specified in config file. \n%r",
-                 connection_config)
-else:
-    connection_config = {
-        'database': 'test',
-        'host': "localhost",
-        'port': 27017,
-        }
```
ericdill, line 18: Is the plan to nuke these defaults?

metadatastore/conf.py
ericdill, line 9: Also, at what point during the install does this file actually get created?

tacaswell, line 9: These files should not be automatically created but provide you a manual way to over-ride the packaged configuration.  Ex our dev accounts have a config which always points to a dev server so we have less chance of accidentally blitting a production server.

[word salad edit]

metadatastore/conf.py
danielballan, line 18: Yes, Stuart wants MDS and FS to raise rather than risk putting real data in a "default" location that is not being intentionally maintained. In practice, there will still be a default, but it will come from our conda packaging, not from MDS and FS themselves.

metadatastore/conf.py
danielballan, line 9: @ericdill 'metadatastore' is on the next line.

danielballan:
Force-pushed to fix the logic, add `_CONDA_ETC`, and configure environmental variables on Travis.

metadatastore/conf.py
cowanml, line 9: +1 on splitting the package defaults from the config file exposed to users.

It's nice when an installer:
      - checks to see if a config file already exists
      - installs a well commented config file, with an additional suffix
        if the file already exists (ala. ".rpmnew" for rpm packages)
      - ideally no default passwords for people to forget to change;
        setup with an init script instead

metadatastore/conf.py
tacaswell, line 8: Ah, this is root `etc`, this needs a leading '/' otherwise to look local to the cwd.

metadatastore/conf.py
```diff
@@ -3,18 +3,31 @@
...
-    logger.debug("Using default db connection. \n%r",
-                 connection_config)
+
+def load_configuration(name, prefix, fields):
```
tacaswell, line 7: This needs to be documented + precedence order

metadatastore/conf.py
```diff
@@ -3,18 +3,31 @@
...
+                ]
+    if 'CONDA_ETC_' in os.environ:
+        filenames.insert(0, os.path.join(os.environ['CONDA_ETC_'], name + '.yml'))
+
+    config = {}
+    for filename in filenames:
+        if os.path.isfile(filename):
+            with open(filename) as f:
+                config.update(yaml.load(f))
+            logger.debug("Using db connection specified in config file. \n%r",
```
tacaswell, line 20: logging should hold off until the final file is read.

ericdill, line 20: why?

metadatastore/conf.py
ericdill, line 18: If mds/fs are not going to install config files by default, we should have examples of what they should look like here: http://nsls-ii.github.io/install.html, right?

PR #125, opened 2015-03-31
BUG : fix up configuration parsing

tacaswell:
un-breaks master


danielballan:
:+1: for better checking of missing fields

PR #128, opened 2015-04-01
DOC: Provide documentation for defining mds config

### LINKED ISSUE #127: Provide an example config file somewhere

ericdill:
- closes #127

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,39 @@
 # metadatastore
 NSLS2 Beamlines metadatastore prototype implemented in MongoDB.
 
-Starting Ophyd
---------------
-export IPYTHONDIR=/epics/xf/23id/ipython/
 
-ipython --profile=xf23id1
+## metadatastore configuration
+
+Metadatastore requires the following configuration information:
+
+```python
+database: metadatastore
+port: 27017
+host: 127.0.0.1
+timezone: US/Eastern
+```
+
+where 
+
+ - `metadatastore` can be any valid string 
+ - `127.0.0.1` can be any IP/DNS name
+ - `US/Eastern` can be any [of these timezone strings] (https://www.vmware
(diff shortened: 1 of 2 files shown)


ericdill:
I was depending on the built-in configuration in metadatastore and filestore. Without this information provided somewhere in the library, I fear that it is exceedingly difficult to stand up a filestore/metdatastore combo on ones local machine.

cowanml:
I strongly encourage:
- defaults in a file somewhere in the package, not intended to be edited
- installing a heavily commented example config wherever the normal
   system level place intended for customizing the config is.
  
   Look at /etc/ssh/sshd_config or something.  After a fresh install
   it basically does nothing, all the entries are commented, but many
   commonly edited settings are documented with commented entries for
   the default settings and nice text descriptions.

Makes it easy to tweak common things, just uncomment and customize the value...
but if you totally hose the file or it's blown away, the actual defaults
are somewhere else safe.

There should be no default passwords anywhere.  Should be
nonfunctional until an init routine is run which prompts for a
password and configures the server and application with it.

tacaswell:
Fortunately are configuration files are not that complicated (and hopefully never will be!)

@stuwilkins Requested that at least MDS and FS _not_ have valid default behavior to force there always to be a human in the loop to set them up (so we never accidentally start writing production data to a localhost mongo server by mistake).  Same idea as no default passwords (well, in pyOlog there is), but as we currently only have trusted links this seems like a reasonable proxy.

There are conda packages (ex  https://conda.nsls2.bnl.gov/CSX/csx_configuration) that will install the beamline specific configuration to the right place.  Should probably also add a `dev_mongo_config` which points everything at local host to make our lives a bit easier.

I am not a fan of system-wide config installation (not even of user-wide config unless it can't be helped). Using system-level config defeats the purpose of sandboxing and eventually we are going to end up deploying to systems where the user may not have root (or have cranky users like me that do not want to).

ericdill:
bump

PR #131, opened 2015-04-13
BLD : update meta.yaml

tacaswell:


conda-recipe/meta.yaml
```diff
@@ -26,6 +26,9 @@ requirements:
     - pandas
 
 test:
+  requires:
+    - nslsii_dev_configuration
+
```
ericdill: add an "about" section

PR #133, opened 2015-04-22
fix + tests for dst transition behavior and human friendly times

cowanml:
The friendly times replacement didn't handle daylight savings transitions quite right... fixed and added tests.  Should be the same as the pandas implementation now.

### DIFF
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -445,7 +445,9 @@ def _normalize_human_friendly_time(val):
         return val
 
     if val.tzinfo is None:
-        val = zone.localize(val)
+        # is_dst=None raises NonExistent and Ambiguous TimeErrors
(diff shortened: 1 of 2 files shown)


tacaswell:
Do I want to ask how badly you have been bit by DST related issues to have remembered this?

tacaswell:
Also : https://stackoverflow.com/questions/6841333/why-is-subtracting-these-two-times-in-1927-giving-a-strange-result

cowanml:
On Tue, 21 Apr 2015, Thomas A Caswell wrote:

> Do I want to ask how badly you have been bit by DST related issues to have
> remembered this?

nah, just something from skimming the pytz docs didn't sit well.  Either
that, or I was burned bad enough to have suppressed the memory :)

PR #134, opened 2015-04-24
PRF/API mongoengine speedup

tacaswell:
Don't re-build the EventDescriptor and DataKeys documents for every event.

A subtle API change is that now all of the events for the same event descriptor share the same python object of that descriptor if they are returned from the same call to `find/fetch_events`

### DIFF
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -9,7 +9,7 @@ from .document import Document
 import datetime
 import logging
 from metadatastore import conf
-from mongoengine import connect
+from mongoengine import connect,  ReferenceField
 import mongoengine.connection
 
 import datetime
@@ -367,6 +367,18 @@ class _AsDocument(object):
 
     def __call__(self, mongoengine_object):
         return Document.from_mongo(mongoengine_object, self._cache)
+
+
+class _AsDocumentRaw(object):
+    """
+    A caching layer to avoid creating reference objects for _every_
+    """
+    def __init__(self):
+        self._cache = dict()
+
+    def __call__(self, name, input_dict, dref_fields):
+
+        return Document.from_dict(name, input_dict, dref_fields, self._cache)
 
 
 def _format_time(search_dict):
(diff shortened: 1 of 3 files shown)


tacaswell:
@klauer This fixes https://github.com/NSLS-II/dataportal/issues/169

```
In [5]: %timeit list(db.fetch_events(db[512]))

1 loops, best of 3: 3.76 s per loop

```

Modulo a failing test + some spam suppression.

klauer:
That's a huge improvement - thanks for taking care of it so quickly!

metadatastore/commands.py
```diff
@@ -358,8 +358,28 @@ class EventDescriptorIsNoneError(ValueError):
 # DATABASE RETRIEVAL ##########################################################
 
 # TODO: Update all query routine documentation
-def _as_document(mongoengine_object):
-    return Document.from_mongo(mongoengine_object)
+class _AsDocument(object):
+    """
+    A caching layer to avoid creating reference objects for _every_
```
danielballan, line 363: "for _every_" trails off here and again below

metadatastore/document.py
```diff
@@ -102,14 +116,85 @@ def from_mongo(cls, mongo_document):
...
             document[field] = attr
-            # For debugging, add a human-friendly time_as_datetime attribute.
-            if 'time' in document:
-                document.time_as_datetime = datetime.fromtimestamp(
-                        document.time)
+        # For debugging, add a human-friendly time_as_datetime attribute.
+        if 'time' in document:
+            document.time_as_datetime = datetime.fromtimestamp(
+                    document.time)
+        return document
```
danielballan, line 140: Good catch.

danielballan:
See line notes. None crucial

metadatastore/commands.py
tacaswell, line 363: I was avoiding caching the rest of the sentence.

danielballan, line 363: :+1:

PR #135, opened 2015-04-29
FIX: A test was bleeding data into the main db.

danielballan:


tacaswell:
I do not understand why this was causing bleeding...

tacaswell:
attn @cowanml This should fix the problem with the default database being created.

cowanml:
I've confirmed this eliminates the 'test' db polution.
thanks!

on a vaguely related note...
I'd have a slightly warmer+fuzzy feeling if utils/testing.py had
   db_name = 'mds_disposable_test_db__{0}'.format(str(uuid.uuid1()))

or something instead of just the very generic, bare uuid4:
   db_name = str(uuid.uuid4())

From the docs, uuid4 is purely random, no time/host involved like for
uuid1.  The odds of inadvertently blowing away a db someone cares
about before the heat death of the universe might by pretty slim in
theory, but...  I can all too easily imagine something screwy
happening with random # seeding or something... makes my gut grumble
slightly.  Or maybe it's it's just dinner time? :)

-matt

tacaswell:
It probably is just dinner time, but I am susceptible to such concerns.  I
also ended up with a large number of uuid named dbs on multiple machines
from debugging testing/killing tests part of the way through so having a
common prefix seems like a good idea to me.

PR #139, opened 2015-04-30
ENH: Add count_events, a stopgap.

danielballan:
Now that `find_events` returns a generator, there is no way to know how many Events are there without looping through them all. This add a new function `count_events` that uses mongo to count the size of the query results without actually building the Event documents.

_This is a temporary solution._ In the long term, it will likely be replaced by one of the following:
- counting each Event stream at the end of a scan and storing those tallies in the RunStop
- making `find_events` return a special pims-like object, a generator that knows its own length

### DIFF
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -638,6 +638,29 @@ def find_events(**kwargs):
 
 
 @_ensure_connection
+def count_events(**kwargs):
+    """Like find_events, but it returns the count, not the actual Events.
+
+    This function is likely to be deprecated in this near future. Instead,
+    find_events will return an iterator that has a length.
+    """
+    # Some user-friendly error messages for an easy mistake to make
+    if 'event_descriptor' in kwargs:
+        raise ValueError("Use 'descriptor', not 'event_descriptor'.")
+    if 'event_descriptor_id' in kwargs:
+        raise ValueError("Use 'descriptor_id', not 'event_descriptor_id'.")
+
+    _format_time(kwargs)
+    try:
+        kwargs['descriptor_id'] = kwargs.pop('descriptor').id
+    except KeyError:
+        pass
+    _normalize_object_id(kwargs, '_id')
+    _normalize_object_id(kwargs, 'descriptor_id')
+    return len(Event.objects(__raw__=kwargs))
+
+
+@_ensure_connection
 def find_last(num=1):
     """Locate the last `num` RunStart Documents
 
--- a/metadatastore/api.py
+++ b/metadatastore/api.py
@@ -1,7 +1,7 @@
 # Data retrieval
 from .commands import (find_events, find_event_descriptors,
                        find_beamline_configs, find_run_starts, find_last,
-                       find_run_stops)
+                       find_run_stops, count_events)
 # Data insertion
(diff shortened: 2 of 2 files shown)


metadatastore/commands.py
```diff
@@ -638,6 +638,29 @@ def find_events(**kwargs):
...
+        raise ValueError("Use 'descriptor_id', not 'event_descriptor_id'.")
+
+    _format_time(kwargs)
+    try:
+        kwargs['descriptor_id'] = kwargs.pop('descriptor').id
+    except KeyError:
+        pass
+    _normalize_object_id(kwargs, '_id')
+    _normalize_object_id(kwargs, 'descriptor_id')
+    return len(Event.objects(__raw__=kwargs))
```
tacaswell, line 660: I suspect that `Event.objects(__raw__=kwargs).count()` will be faster as it will skip the making all of the mongoengine objects step (which were the source of the major performance hit at HXN).

ericdill, line 660: Doing this on the HXN scan that used to take a long time is pretty snappy already. (Note that this is done with hash 4254230070efd1c897bc0763dae4895aa34068ad). 

``` python
In [1]: from metadatastore.commands import *
/home/edill/miniconda/envs/ophyd/lib/python2.7/site-packages/pytz/__init__.py:29: UserWarning: Module bson was already imported from /home/edill/miniconda/envs/ophyd/lib/python2.7/site-packages/bson/__init__.pyc, but /home/edill/miniconda/envs/ophyd/lib/python2.7/site-packages/mongoengine-0.8.7-py2.7.egg is being added to sys.path
  from pkg_resources import resource_stream

In [2]: rs = next(find_run_starts(scan_id=512))

In [3]: evd = next(find_event_descriptors(run_start_id=rs.id))

In [4]: %timeit find_events(descriptor_id=evd.id)
The slowest run took 26.67 times longer than the fastest. This could mean that an intermediate result is being cached 
1000 loops, best of 3: 288 µs per loop

In [5]: pwd
Out[5]: u'/home/edill'

In [6]: import os

In [7]: os.uname()
Out[7]: 
('Linux',
 'xf03idc-ws2',
 '3.2.0-4-amd64',
 '#1 SMP Debian 3.2.65-1+deb7u2',
 'x86_64')
```

ericdill:
I'm :+1: on this

metadatastore/commands.py
danielballan, line 660: Actually, the mongoengine docs [suggest that `len` is better](https://mongoengine-odm.readthedocs.org/guide/querying.html#counting-results). I assume it passes through to `.count()`, but it's more Pythonic.

ghost:
count() in all mongo drivers run the query that generated the specific cursor you're counting and iterates through all of its objects in order to figure out how many elements are part of the cursor. There is no way around this since mongo is not transactional in the traditional sense. len() is done on the client side  and cursor takes place on the server side. Avoiding a round trip to database server is more desirable, that's why mongoengine documentation suggests using len.

danielballan:
I see. Well, my original motivation for this is gone. I had forgotten something clever I had already done to work around this. We can merge or let this sit and wait for a better idea -- no opinion here.

PR #140, opened 2015-05-14
API: Interface to MDS is uids, not mongoengine documents.

danielballan:
We are moving toward a model where ophyd builds dictionaries matching the established document spec and emits them to generic consumers. MDS is merely one consumer, and other consumers could be alongside of or in place of it. Thus, the MDS API should not deal in mongoengine documents. It should accept and return uids.

Before:

```
run_start_document = insert_run_start(...)
event_descriptor_document = insert_event_descriptor(run_start_document, ...)
```

After:

```
run_start_uid = insert_run_start(...)
event_descriptor_uid = insert_event_descriptor(run_start_uid, ...)
```

This PR does not touch `odm_templates.py` and therefore does not require a database migration.

More Details:
- The argument for the run start uid is simply `run_start` not `run_start_uid`, so that dictionaries matching the document spec (which has `run_start`, not `run_start_uid`) can be unpacked into the function: `insert_event_descriptor(**my_descriptor_dict)`

### DIFF
--- a/metadatastore/commands.py
+++ b/metadatastore/commands.py
@@ -126,8 +126,8 @@ def insert_run_start(time, scan_id, beamline_id, beamline_config, uid=None,
     beamline_id : str
         Beamline String identifier. Not unique, just an indicator of
         beamline code for multiple beamline systems
-    beamline_config : metadatastore.odm_temples.BeamlineConfig
-        Foreign key to beamline config corresponding to a given run
+    beamline_config : str
+        uid of beamline config corresponding to a given run
     uid : str, optional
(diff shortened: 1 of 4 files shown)


ghost:
As @danielballan and I discussed, this is a small change that makes us less backend technology dependent(not changing anything, as a good cs practice) and it is easier to work with native types compared to mongoengine specific docs, which sometimes change the behavior and break our framework(as in changes between 0.8.7 and 0.9.x). 
:+1:  from me as soon as it passes the tests :hammer:

ericdill:
Closing, because these commits are in #142

PR #142, opened 2015-05-14
WIP: Uids for all

ericdill:
Supersedes #140

~~Note that this is not yet ready for merge, I wanted a place to compare all the changes while I finish the dev on this.~~


ericdill:
Ready for review

ericdill:
Ah, shoot. I need to add `uid` to the index on mongo

ericdill:
What in the hell are these python 3 bugs??

tacaswell:
So, that is interesting.  Writing  an int literal with a leading 0 makes it base 8 (?!).

``` python
>>> 01
1
>>> 02
2
>>> 03
3
>>> 04
4
>>> 05
5
>>> 06
6
>>> 07
7
>>> 08
  File "<stdin>", line 1
    08
     ^
SyntaxError: invalid token
>>> 010
8
>>> 011
9
>>> 012
10
>>> 013
11
>>> 030213
12427
>>> 

```

This is justifiable confusing and was removed in 3 See https://www.python.org/dev/peps/pep-3127/

metadatastore/examples/sample_data/multisource_event.py
```diff
@@ -49,12 +51,24 @@ def run(run_start=None, sleep=0):
...
         events.append(event)
-
     return events
 
 
 if __name__ == '__main__':
-    run()
+    import metadatastore.api as mdsc
+    blc_uid = mdsc.insert_beamline_config({}, time=0.)
+    run_start_uid = mdsc.insert_run_start(scan_id=02032013,
```
tacaswell: remove the leading `0`

metadatastore/examples/sample_data/temperature_ramp.py
```diff
@@ -34,20 +36,37 @@ def run(run_start=None, sleep=0):
...
         events.append(event)
 
     return events
 
 
 if __name__ == '__main__':
-    run()
+    import metadatastore.api as mdsc
+    blc_uid = mdsc.insert_beamline_config({}, time=0.)
+    run_start_uid = mdsc.insert_run_start(scan_id=030213,
```
tacaswell: remove the leading `0`

metadatastore/test/test_commands.py
```diff
@@ -1,94 +1,80 @@
...
 
 def setup():
     mds_setup()
-    global blc
+    global blc_uid, run_start_uid, document_insertion_time
+    document_insertion_time = ttime.time()
     temperature_ramp.run()
-    blc = mdsc.insert_beamline_config({}, ttime.time())
+    blc_uid = mdsc.insert_beamline_config({}, time=document_insertion_time)
+    run_start_uid = mdsc.insert_run_start(scan_id=030213,
```
tacaswell: remove leading `0`

metadatastore/test/test_commands.py
```diff
@@ -100,165 +86,163 @@ def test_ev_desc():
...
-
-
-def test_run_start():
+def test_insert_run_start():
     time = ttime.time()
     beamline_id = 'sample_beamline'
-    yield _run_start_tester, time, beamline_id, 42
+    scan_id = 42
+    custom = {'foo': 'bar', 'baz': 42,
+            'aardvark': ['ants', 3.14]}
```
tacaswell, line 119: Dora: "ant pie! Yum!"

metadatastore/commands.py
```diff
@@ -590,17 +594,16 @@ def find_beamline_configs(**kwargs):
...
 
     Parameters
     ----------
     run_start : metadatastore.document.Document, optional
         The run start to get the corresponding run end for
-    run_start_id : str or ObjectId, optional
-        The unique id generated by mongo for the RunStart Document.
-        NOTE: If both `run_start` and `run_start_id` are provided,
-              `run_start.id` supersedes `run_start_id`
+    run_start_uid : str
```
tacaswell: ~~Please add note about precedence back~~ :sheep: nm, can't read.

tacaswell: sorry for the noise

metadatastore/commands.py
```diff
@@ -627,11 +630,19 @@ def find_run_stops(**kwargs):
     run_stop : iterable of metadatastore.document.Document objects
     """
     _format_time(kwargs)
+    if run_start and run_start_uid and run_start.uid != run_start_uid:
```
tacaswell: This logic should probable get pulled out into a function as it shows up 3x.

ericdill:
@danielballan can I punt this back to you? Only a few changes left and I'm in the car all day today.

metadatastore/commands.py
```diff
@@ -627,11 +630,19 @@ def find_run_stops(**kwargs):
...
+        raise ValueError(
+            "If you insist on specifying both the run_start and the "
+            "run_start_uid kwargs, please ensure that they are actually "
+            "the same value...\nrun_start.uid={}\nrun_start_uid={}"
+            "".format(run_start, run_start_uid))
```
danielballan: I love the snark here.

danielballan:
Yes. Thanks for tapping in.

metadatastore/commands.py
ericdill: good call; bab2e81

ericdill:
@danielballan @tacaswell I think this should be good to go. Bugs (#143 in particular) are fixed and tests are passing

metadatastore/commands.py
```diff
@@ -157,6 +157,7 @@ def insert_run_start(time, scan_id, beamline_id, beamline_config, uid=None,
     if project is None:
         project = ''
 
+    beamline_config = BeamlineConfig.objects(uid=beamline_config).first()
```
tacaswell: Maybe use `BeamlineConfig.get(uid=beamline_config)` which is marginally faster (maybe?) and will raise if there is any number of results other than 1.

metadatastore/commands.py
```diff
@@ -277,6 +279,7 @@ def insert_event_descriptor(run_start, data_keys, time, uid=None,
     if custom is None:
         custom = {}
     data_keys = format_data_keys(data_keys)
+    run_start = RunStart.objects(uid=run_start).first()
```
tacaswell: ditto

metadatastore/commands.py
```diff
@@ -316,13 +318,15 @@ def insert_event(event_descriptor, time, data, seq_num, uid=None):
...
-    # mostly here to notify ophyd that an event descriptor needs to be created
+    # Allow caller to beg forgiveness rather than ask permission w.r.t
+    # EventDescriptor creation.
     if event_descriptor is None:
         raise EventDescriptorIsNoneError()
 
     if uid is None:
         uid = str(uuid.uuid4())
 
+    event_descriptor = EventDescriptor.objects(uid=event_descriptor).first()
```
tacaswell: ditto

metadatastore/commands.py
```diff
@@ -494,6 +498,23 @@ def _normalize_object_id(kwargs, key):
...
+        if document_uid and document.uid != document_uid:
+            raise ValueError(
+                "If you insist on specifying both the document and the "
+                "uid kwargs, please ensure that they are actually "
+                "the same value...\ndocument.uid={}\n*_uid={}"
+                "".format(document, document_uid))
+        document_uid = document.uid
+
+    if document_uid:
+        document = document_cls.objects(__raw__={'uid': document_uid})[0]
```
tacaswell: use `.get` here too

metadatastore/examples/sample_data/multisource_event.py
tacaswell: bump on this.

metadatastore/odm_templates.py
```diff
@@ -23,7 +23,7 @@ class BeamlineConfig(DynamicDocument):
     config_params = DictField(required=False, unique=False)
     uid = StringField(required=True, unique=True)
     time = FloatField(required=True)
-    meta = {'indexes': ['-_id'], 'db_alias': ALIAS}
+    meta = {'indexes': ['-_id', '-uid'], 'db_alias': ALIAS}
```
tacaswell, line 30: good catch on these!

metadatastore/commands.py
```diff
@@ -126,8 +126,8 @@ def insert_run_start(time, scan_id, beamline_id, beamline_config, uid=None,
     beamline_id : str
         Beamline String identifier. Not unique, just an indicator of
         beamline code for multiple beamline systems
-    beamline_config : metadatastore.odm_temples.BeamlineConfig
-        Foreign key to beamline config corresponding to a given run
+    beamline_config : str
+        uid of beamline config corresponding to a given run
```
ericdill: Note to self, this docstring is wrong

ericdill: Oh wait nevermind, I was not reading closely

ericdill: However, these (and similar params in other insert functions) should be renamed to `beamline_config_uid`

metadatastore/commands.py
ericdill: `.get()` is not a valid method

http://nbviewer.ipython.org/gist/ericdill/ca047302c2c1f1865415

cowanml: > In metadatastore/commands.py:
> 
> > @@ -157,6 +157,7 @@ def insert_run_start(time, scan_id, beamline_id, beamline_config, uid=Non
> > e,
> >      if project is None:
> >          project = ''
> > -    beamline_config = BeamlineConfig.objects(uid=beamline_config).first()
> 
> .get() is not a valid method
> 
> http://nbviewer.ipython.org/gist/ericdill/ca047302c2c1f1865415

I think he meant:

```
 BeamlineConfig.objects(uid=beamline_config).get()
```

ericdill: Oh. Thanks...  I updated the notebook with `%timeit` results for a number of ways to search for objects in mds.  

Interesting results: 
1. `.get()` is slower than `.first()` which is slower than `[0]`.  
2. `RunStart.objects(__raw__={'uid':uid})` is faster than `RunStart.objects(uid = uid)`

metadatastore/odm_templates.py
tacaswell, line 30: On further consideration, we get `uid` indexed automatically because we are constraining it to be unique.

metadatastore/commands.py
tacaswell: ok then.  `[0]`  it is, but it probably should also get a comment.

I guess the use-case for `get` is in cases where you have an optional field where the non-null entries should be unique (and you would like to know when that is not true).  Looking at the code, `get` does a bunch of extra work to raise in the case were there is more than one result.  We are ok with out this check because the field is enforced to be unique by mongo already.

danielballan: +1 for benchmarking!  A gold star for you hahaha

metadatastore/commands.py
danielballan: I see the argument for appending `_uid` to the name, but I think it actually the wrong choice. Review my original comments in #140.

ericdill: My only issue is that the kwargs to `insert_*` and `find_*` are not symmetric. On the `find_*` side, `run_start` refers to the whole mds document, while `run_start_uid` is the uid of the run_start. On the insert_\* side, `run_start` is the uid of the run_start.  I guess it just feels a little strange, but other than that I'm not sure I can come up with a compelling argument

ericdill:
scope_creep 20pic

metadatastore/commands.py
danielballan: Hmm. That bothers me too. I guess we could change `find_*` to access uids or Documents. There will never be any ambiguity there. Let's think on this.

metadatastore/odm_templates.py
ghost, line 30: Yup, meta is usually for sparse indexing

ericdill, line 30: Then I don't understand why `id` is indexed on all of the documents

ghost, line 30: Because the default indexing of mongoengine is ascending order. We want to ensure _id are done in descending order for quicker search. _id is the primary key for retrieving the actual documents in the backend. Specifying that “-“ makes the queries a little bit faster with growing document size.

On May 26, 2015, at 9:24 AM, Eric Dill <notifications@github.com<mailto:notifications@github.com>> wrote:

Then I don't understand why id is indexed on all of the documents

metadatastore/commands.py
ghost: I agree with Eric. We do need to get rid off this ambiguity. At this point, we all know the schema very well so it is fine for us but for another person using this library for the first time, it is quite confusing. We could simply upgrade to the new mongoengine library that I belive fixes this issue.

ghost:
This looks good to go! Thanks @ericdill !

metadatastore/test/test_commands.py
danielballan, line 119: @doraf makes me laugh.

metadatastore/odm_templates.py
ericdill, line 30: Ah. Well then I would imagine we want uid indexed the same way id is to get the same benefits, right?

ericdill:
5b09484 addresses the discussion of asymmetric `insert_*` and `find_*` functions. Assuming travis passes, this ~~should be~~ is ready for final review and/or merge

metadatastore/odm_templates.py
tacaswell, line 30: does not matter for the uids because they are random.  Objectids are prefixed by a time stamp so they are monotonically increasing.

ghost, line 30: Of course it matters. The daemon has a priority buffer that keeps the most recent stuff that is denoted by "-". Eric does the right thing by ensuring it is indexed properly. Also, meta has tons of neat features we might use in the future.

PR #146, opened 2015-05-26
Nice HTML representations of Headers from Data Broker

danielballan:
Demo: http://nbviewer.ipython.org/gist/danielballan/95dd38e97f12ddce7942


ericdill:
:+1: . passes py2/py3 tests locally. merging.

PR #149, opened 2015-05-27
Numpy repr

danielballan:
Two things:
- Tweaking examples for DAMA demo
- Special casing numpy array in Header's HTML repr so that it does not break out into a table.


ericdill:
Passes tests, I assume you know what you are doing here, so :+1:

PR #150, opened 2015-05-29
Mds corrections

ericdill:
When complete, this PR will introduce the ability to update metadatastore documents via a new `Correction`s collection (sure is a tongue twister!) and an `update` function in metadatastore.  A `Correction`s document is a `DynamicDocument` with three fields:
1. `uid`: The unique identifier for one of the other documents (`RunStart`, `RunStop`, `EventDescriptor`, `BeamlineConfig`)
1. `corrections_uid`: A unique string identifier for one `Correction`s document. Defaults to `uuid4` if none is provided
1. `time`: The unix epoch time that the document was saved

Note that `Event` documents cannot be updated.  `update(document)` will raise if `isinstance(document, Event)`

This PR introduces a new kwarg into the `find_*` functions, initially called `use_newest_correction=True` which is a boolean flag that will:
- if True: instruct the `find_*` function to go and grab the most recent correction for that document (based on its `uid`) 
- if False: use the original document.

This PR introduces a new function `find_corrections` that will retrieve all corrections for a document based on either the original document's `uid` or the `correction_uid` of a specific correction.

This PR requires test coverage before it can be merged.
## Todo
- [x] full test coverage
- [x] update(RunStart) works
- [x] update(RunStop) works
- [x] update(EventDescriptor) works
- [x] update(BeamlineConfig) works
- [x] update(Event) raises (https://github.com/ericdill/metadataStore/commit/bf587918e1924d78d4fbfa2f5b21a7f3d3f9b649)
- [x] find_corrections(uid=uid) returns 0-N correction documents
- [x] find_corrections(correction_uid=correction_uid) returns 0-1 correction documents
- [x] find_\* functions correctly grab the newest document stack when asked
- [x] find_event correctly grabs the newest EventDescriptor when asked
- [x] find_event does not recursively check for new documents. If event.descriptor is an EventDescriptor document then [this code path](https://github.com/NSLS-II/metadatastore/pull/150/files#diff-86f2983c849d2ca24f0c14643a623d33R648) will short-circuit and not check for a new `RunStart`.
- [x] Add a field to the `Correction` document that keeps track of which user account made the change
## Demo

http://nbviewer.ipython.org/ff3540dbe06073a673ec
## Known use cases
1. RunStart points to the wrong beamline config (https://github.com/ericdill/metadataStore/commit/6b9792e0e2d1af2c41470a7e06b8ad6c30fb9e12)
   1. Insert a new beamline configuration with `insert_beamline_configuration`. This adds to the `BeamlineConfig` collection
   2. Update RunStart to point to new beamline configuration
   3. call `update` on the updated RunStart document. This adds to the `Correction` collection.
2. Annotate `RunStart`, `EventDescriptor`, `RunStop` documents with new tags (https://github.com/ericdill/metadataStore/commit/6b9792e0e2d1af2c41470a7e06b8ad6c30fb9e12)
   1. Edit `run_start`
   2. call `update(run_start). This adds to the`Correction`collection. Future calls to`find_run_starts(uid=run_start.uid)`will return the tagged run_start document. If you want the original document,`find_run_starts(uid=run_start.uid, use_newest_correction=False)` will return it
3. EventDescriptor has wrong shape for data key
   1. Edit `event_descriptor`
   2. call `update(event_descriptor)`. This adds to the `Correction` collection. Future calls to `find_event_descriptors(uid=event_descriptor.uid)` will return the corrected event descriptor. If you want the original document, `find_event_descriptors(uid=event_descriptor.uid, use_newest_correction=False)` will return it
4. BeamlineConfig contains inaccurate information
   1. Similar to the first point. Update the beamline config. Save it. Update all RunStart's which were using that beamline config.


tacaswell:
can you nuke and ban the ipython notebook checkpoints?

metadatastore/commands.py
```diff
@@ -48,6 +57,11 @@ def db_disconnect():
 
 
 def db_connect(database, host, port):
+    try:
```
tacaswell: Why did this get added?

tacaswell:
and would it be possible to get that object graph as any kind of vector graphic?

metadatastore/commands.py
```diff
@@ -541,13 +550,137 @@ def _get_mongo_document(document, document_cls):
     mongo_document = document_cls.objects(__raw__={'uid': document})[0]
     return mongo_document
 
+def _dereference_uid_fields():
+    return
 
 @_ensure_connection
-def find_run_starts(**kwargs):
+def find_corrections(newest=False, dereference_uids=True, **kwargs):
```
tacaswell, line 559: `derefernce_uids` is not documented.

metadatastore/commands.py
```diff
@@ -541,13 +550,137 @@ def _get_mongo_document(document, document_cls):
...
+            document = Document(document, klass.__name__)
+            if newest and 'beamline_config' not in ref:
+                corrected_document_generator = find_corrections(
+                    uid=document['uid'])
+                try:
+                    document = next(corrected_document_generator)
+                except StopIteration:
+                    pass
+            _dereference_reference_fields(document, newest=newest)
+            if ref.endswith('_id'):
```
tacaswell, line 645: Can you add a comment about why this is being done so we remember next time we read this.

metadatastore/commands.py
```diff
@@ -541,13 +550,137 @@ def _get_mongo_document(document, document_cls):
...
+        if field:
+            # then field is either an ObjectId or a uid, or the object model
+            # is borked...
+            if isinstance(field, ObjectId):
+                # find the original document
+                kwargs = {'_id': field}
+            elif isinstance(field, six.text_type):
+                # it better be a uid
+                kwargs = {'uid': field}
+            else:
```
tacaswell, line 631: Can you add a comment that this is to catch cases of already de-referenced documents.

metadatastore/commands.py
```diff
@@ -541,13 +550,137 @@ def _get_mongo_document(document, document_cls):
...
+
+    Returns
+    -------
+    as_document : list
+        List of mongo dictionaries
+    """
+    _normalize_object_id(kwargs, '_id')
+    _format_time(kwargs)
+    with no_dereference(DocumentClass) as DocumentClass:
+        docs = DocumentClass.objects(__raw__=kwargs).order_by('-_id')
```
tacaswell, line 674: do we want to order by time instead?

metadatastore/commands.py
```diff
@@ -671,24 +818,34 @@ def find_run_stops(run_start=None, **kwargs):
     -------
     run_stop : iterable of metadatastore.document.Document objects
     """
-    _format_time(kwargs)
-    # get the actual mongo document
     if run_start:
-        run_start = _get_mongo_document(run_start, RunStart)
-        kwargs['run_start_id'] = run_start.id
+        run_start_uid = _get_uid(run_start)
```
tacaswell, line 836: Why is the behavior of returning `run_start` the right thing to do here if `run_start` does not have a uid?

metadatastore/commands.py
```diff
@@ -671,24 +818,34 @@ def find_run_stops(run_start=None, **kwargs):
...
+        mongo_run_start = _find_documents(RunStart, uid=run_start_uid)
+        mongo_run_start = next(mongo_run_start)
```
tacaswell, line 838: This will raise an uncaught exception if the uid is not valid, is that intended?

metadatastore/document.py
```diff
@@ -217,62 +65,52 @@ def __repr__(self):
...
-
     def __str__(self):
-        return self._str_helper(self._name)
+        return _str_helper(self)
 
     def _repr_html_(self):
         return html_table_repr(self)
 
 
+def _normalize(in_val, cache):
```
tacaswell: This function should be deleted

tacaswell:
have you tried running this on the HXN case that was slow?

metadatastore/commands.py
ericdill, line 674: It is the same thing since id has time baked in

metadatastore/commands.py
ericdill: it is half of a solution I had implemented. the solution is no longer useful.

metadatastore/commands.py
ericdill, line 836: `run_start` can be a uid or a document. If `run_start` does not have a uid then it is a uid

metadatastore/commands.py
ericdill, line 838: Perhaps the error that will be raised by this is not the _most useful error_ possible, but seems reasonable to raise an error here.  If you pass in a `RunStart` document or a `run_start_uid` and we cannot find the mongo object that corresponds to that document or uid then something has gone very wrong. 

Feels like the correct thing to do to barf at this point, right?  What error should be raised here instead?

ericdill:
I have not tried the HXN case yet. Do you remember what that scan id was?

metadatastore/commands.py
ericdill, line 631: commented

metadatastore/commands.py
ericdill, line 645: commented

tacaswell:
https://github.com/NSLS-II/dataportal/issues/169

scan 512

ericdill:
re scan 512 on HXN: When I tried to run that last time, I discovered that the channelarchiver had filled up all 11 TB at HXN and 20-something TB at CSX. I'll try it again tomorrow

ericdill:
``` python
In [6]: def time_events(header):
    t0 = ttime.time()
    events = db.fetch_events(header)
    t1 = ttime.time()
    events = list(events)
    t2 = ttime.time()
    print('time for db.fetch_events call = %s s' % (t1-t0))
    print('time to listify the generator = %s s' % (t2-t1))
    return events
   ...: 

In [7]: events = time_events(hdr512)
time for db.fetch_events call = 4.05311584473e-06 s
time to listify the generator = 42.5619840622 s

```

...crap

ericdill:
Here's some profiling output sorted by the total time for each function. I also included the cProfile output saved to disk.  Would you be able to help me sort out why this is slow @tacaswell ?

https://gist.github.com/ericdill/95aa082c489153c0f0a0

help me tacaswell

ericdill:
I suppose this is going to die on the vine. oh well!

PR #152, opened 2015-05-30
REF: Expect new Document format.

danielballan:
There might be more thrashing to come. Don't merge.


danielballan:
P.S. This is not really a _new_ change, just the last step in making MDS a generic "consumer" of Documents issued from ophyd. It's a follow-up on #142.

danielballan:
Updated. Ready to go. Cousin PR in databroker coming momentarily....

PR #154, opened 2015-06-15
MNT: Make test connection more explicit

ericdill:
Note that this is not actually changing any behavior

### DIFF
--- a/metadatastore/utils/testing.py
+++ b/metadatastore/utils/testing.py
@@ -1,18 +1,32 @@
 import uuid
 from metadatastore.api import db_connect, db_disconnect
-
+from metadatastore import conf
+from copy import deepcopy
 
 conn = None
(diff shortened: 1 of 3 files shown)


metadatastore/utils/testing.py
```diff
@@ -1,18 +1,32 @@
...
 
 def mds_setup():
     "Create a fresh database with unique (random) name."
     global conn
+    global old_connection_info
+    old_connection_info = deepcopy(conf.connection_config)
+    conf.connection_config = testing_config
     db_disconnect()
-    conn = db_connect(db_name, 'localhost', 27017)
+    conn = db_connect(testing_config['database'], testing_config['host'],
```
tacaswell, line 23: You can probably drop this and rely on `ensure_connection` to take care of the actual connection for you.

tacaswell, line 23: Maybe, depending on some subtle details of importing that I can not swear to.

PR #155, opened 2015-06-17
API: time_as_datetime -> time_str

danielballan:
Writing a [proof-of-concept data export tool](https://github.com/NSLS-II/suitcase) made it clear that tacking datetime objects onto Documents on their way out was a mistake. HDF5 can't store them natively, and they aren't JSON serializable. They're a Python-ism in a scheme that otherwise avoids them.

We could:
1. Rely on our new Document`__repr__` and `_repr_html_` to take care of making the times readable and remove the extra time field altogether.
2. Use an ISO-format string instead. Notice that changing to strings reveals the subtle problem that datetime casts to int and drop precision:

```
time_as_datetime    2015-06-12 16:57:00.129000
time_as_str 2015-06-12T16:57:00.129079
```

I think (2) is a gentler change from the status quo, and that's what this PR does. We might want to consider (1) though: now that we have nicer user interfaces, maybe we don't need to keep mutating Documents on their way out. Thoughts?

### DIFF
--- a/metadatastore/document.py
+++ b/metadatastore/document.py
@@ -1,9 +1,11 @@
+import re
 import six
 import mongoengine
 from mongoengine.base.datastructures import BaseDict, BaseList
 from mongoengine.base.document import BaseDocument
 from bson.objectid import ObjectId
 from bson.dbref import DBRef
+import time as ttime
 from datetime import datetime
 from itertools import chain
 from collections import MutableMapping, Mapping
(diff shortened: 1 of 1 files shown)


tacaswell:
On the other hand, the datetime objects have nice subtraction and with only a bit of prodding will play nice with plotting.

I think you just have to suck it up and deal with special-casing on the way in and out.

danielballan:
Haha. I already have. Plotting is a good point.

PR #159, opened 2015-06-25
DOC: New line was ruining a link

ericdill:


ericdill:
self-merging, because it is an obvious change

PR #169, opened 2015-09-10
Enh get header

tacaswell:


metadatastore/commands.py
```diff
@@ -1165,3 +1165,46 @@ def find_last(num=1):
         if next(c) == num:
             raise StopIteration
         yield _cache_run_start(rs)
+
+
+def get_header(run_start, allow_no_run_stop):
+    """Make header from a RunStart
+
+    This is simple a
```
danielballan: ...cliffhanger

metadatastore/commands.py
```diff
@@ -1165,3 +1165,46 @@ def find_last(num=1):
...
+
+    Parameters
+    ----------
+    run_start : dict or str
+        The RunStart to build RunRecord from.  Can be either
+        a Document/dict with a 'uid' key or a uid string
+
+    allow_no_run_stop : bool, optional
+        If True, allow the RunRecord to be constructed without
+        a RunStop.  Defaults to False
```
danielballan: No it doesn't.

danielballan: I am +0.5 on default to True.

danielballan:
Fix notes and merge at will.

metadatastore/commands.py
```diff
@@ -1165,3 +1165,46 @@ def find_last(num=1):
         if next(c) == num:
             raise StopIteration
         yield _cache_run_start(rs)
+
+
+def get_header(run_start, allow_no_run_stop):
```
ericdill: Why are we giving a different meaning to "header" in mds and "header" in dataportal?

I.e., 

``` python
db[-1] != mds.get_header(list(mds.find_last()[0]))
```

tacaswell: The plan is to re-name the DP headers to 'RunRecord' or something like that.  This dict is more along the lines of the original thing we meant by 'header'.

ericdill: And why would we want to rename the dataportal Header?

dchabot: On the surface, this sounds like a recipe for confusion...

danielballan: The question is, what db[] should return? The current answer (headers as
implemented) is not a bad one, but it has its problems. I suspect it is
trying to be too many things at once. Let's meet in person next week to
evaluate alternatives.

tacaswell:
@danielballan I assume you don't use this in you DP work?

danielballan:
Yes. I think this is a DP-level abstraction -- not sure it needs to be available in MDS at all. Close if you agree.

tacaswell:
Leaving open as if we do implement a caching layer as discussed with @arkilic it would hold 'headers' + 'tables' so in some sense DP should only _ever_ get headers

tacaswell:
This has been rebased on current master.  Still not convinced that this is not a good idea.

ghost:
I agree that this is a good idea. The latest mongo has some nifty new features that could come handy for this: https://www.mongodb.com/blog/post/joins-and-other-aggregation-enhancements-coming-in-mongodb-3-2-part-1-of-3-introduction

danielballan:
> Leaving open as if we do implement a caching layer as discussed with @arkilic it would hold 'headers' + 'tables' so in some sense DP should only ever get headers

That's reasonably convincing. They don't literally have to be Header objects, but then again I can't think of a good reason to invent a second name for "group of three documents". Shall we let this sit until the caching layer is further along?

ericdill:
Any updates on this?

danielballan:
> Shall we let this sit until the caching layer is further along?

I think yes. :- )

ericdill:
> I think yes. :- )

I thought we had caching in metadatastore already (which is why I bumped this)

tacaswell:
Different caching layer.  The caching currently being done is just to save trips to the DB for information we have already asked for (ex, only get each descriptor once, not once per event) and only hits the Start, Stop, and Descriptor docuents.  The caching we are talking about here is a persistent on-disk cache of all of the documents.  One possible way to serialize this to disk is the de-normalized 'header' in to a YAML and the events as an HDF5 / csv / bunch of numpy files / ... 

I think choices about that (and at which layer we do it) will drive exactly what this will look like. 

It is probably save to let this keep sitting.

ericdill:
> Different caching layer. 

Thanks for clarifying :smile: 

> I think choices about that (and at which layer we do it) will drive exactly what this will look like.

Makes sense.

> It is probably save to let this keep sitting.

Sounds good

PR #174, opened 2015-09-16
Add boltons to conda-recipe

ericdill:


ericdill:
attn @tacaswell the conda recipe does not build for v0.2.0 because boltons was not included in the conda recipe

ericdill:
Once we agree on this conda-recipe, I can put in a PR against conda-prescriptions with the updated recipe

conda-recipe/meta.yaml
```diff
@@ -29,20 +29,17 @@ requirements:
...
 
 test:
   requires:
     - nslsii_dev_configuration
 
-  # Python imports
   imports:
+    # Import all the packages
     - metadatastore
-    - metadatastore.commands
```
tacaswell, line 41: why did these get dropped?

ericdill, line 41: The best rationale that I have for dropping these is that importing the **modules** seems like more trouble then its worth.  The main problem that I have had with setup.py is adding a package in the source and not including it in the setup.py so that it is not part of the package.  It is also easy to get the module names in the source out of sync with the module names in this meta.yaml.  The most recent instance of this is that you redid the `Document` object in `doc.py`, not `document.py` but did not update this meta.yaml file to include this update.

Then again, importing every module will **at least** make sure that we don't have import failures from missing deps?  

I could go either way. 
1. Have all modules in the test imports section
2. Have only packages in the test imports section.

What do you want to do?

ericdill, line 41: Further evidence of us being sloppy with the combination of conda-recipe + setup.py + library thrashing is that in the setup.py we are adding a package that does not exist... `metadatastore.utils`

https://github.com/NSLS-II/metadatastore/blob/master/setup.py#L22

tacaswell, line 41: This is one reason to switch to using `conda build` on travis (assuming we can find a way to cope with the deps).  

I am in favor of 1 and shaming people that forget to update meta.yaml

tacaswell:
And this apparently needs a rebase

conda-recipe/meta.yaml
ericdill, line 41: lmao, ok.  I have added **all** the non-test modules to the meta.yaml.  Let the shaming of @tacaswell commence!

ericdill:
Ok, rebased.  I had added jinja2 to the meta.yaml in the test improvement PR

PR #175, opened 2015-09-16
WIP: Run test suite in conda build

ericdill:


ericdill:
### Benefits to running the test suite from within the conda recipe
- **We don't build and upload binaries whose tests fail...**

### Problems with this approach
- Need to have mongodb running
  - conda-build locations where this is not a problem
    - If the binaries are going to be built locally, that is not a problem
    - If the binaries are going to be built inside docker, running mongodb is not a problem
    - If the binaries are going to be built on our internal conda server via binstar workers then running mongodb on our local apple mac pro alongside the binstar workers is not a problem
  - conda-build locations where this is a problem
    - If the binaries are going to be built on anaconda.org via automagical `binstar build submit <path_to_binstar_yaml>` then we are (probably?) not going to be able to run this test suite

Thoughts?

[edit] bad english

ericdill:
this is a bad idea, re discussion on a walk to lunch

PR #176, opened 2015-09-18
WIP: conda-build on travis

ericdill:


ericdill:
attn @danielballan @tacaswell conda-build is functional on travis!

.travis.yml
```diff
@@ -1,37 +1,49 @@
...
+sudo: false
+
+services:
+  - mongodb
+
+matrix:
+  include:
+    - python: 2.7
+    - python: 3.4
+    - python: 3.5
```
danielballan: I think 3.5 is here by mistake.

danielballan: Better yet, remove the one under `allow_failures` and insist of 3.5 support!

ericdill: it is here because some of the deps are not yet built on 3.5 and I didn't feel like building them yet.

ericdill: buuuuut I am building them now. Hold tight.

ericdill:
attn @danielballan @tacaswell This should go in soon.  It adds the following:
- faster installs because conda is heavily used
- testing on python 3.5
- conda build with the current conda-recipe in the repo

tacaswell:
The tests never get run?

tacaswell:
Oh, I see.  Might be worth building the conda package, installing in and then running the tests off of that.

ericdill:
Testing a conda-built package requires a pretty major overhaul of our test suite.  Right now, nose uses the source code to auto-discover the tests. Here is one solution that I can implement pretty easily:

Move the tests from metadatastore/metadatastore/test to metadatastore/test. Then on travis I can rm -r the metadatastore directory which will force the tests to use the conda installed version.  

Or we can punt on this for now and just let the conda-build serve as a (reasonably effective) smoketest that makes sure that all the modules are importable.

ericdill:
Also note: travis is passing but this should not be merged right now

ericdill:
Also, I want to punt on making travis test the conda-build for now because I think I can do this more sustainably by using conda-skeletor to find all the deps/testing files for me

tacaswell:
this seems reasonable to me

ericdill:
cff1873 **should** upload the built packages to lightsource2-dev. It's a little hard to test since it only works when travis executes on the master branch. Environmental variable are disabled on builds for pull requests.

tacaswell:
merged to see what happens ;)

@ericdill as a side note, the encrypted envs are available on pushes to any branch on upstream so you can work on this sort of thing by pushing branches up stream (a perfectly good reason to break the rule against pushing rouge branches to upstream). The logic is that any old person can open a PR, only people with commit rights can push to upstream.

ericdill:
Looks like it worked! Sort of. 2.7 and 3.4 got built, haven't yet looked in
to why 3.5 did not build.

http://anaconda.org/lightsource2-dev/metadatastore/files

Good point about other branches Tom thanks.  In production though, I should
add a check to make sure we are on the master branch before uploading to
lightsource2-dev

PR #180, opened 2015-09-29
Add 'scan_type' to run start.

ericdill:
I want this in the run start because it is getting put in by bluesky.
It is pretty helpful to have the same info in the sample data generators that
we expect to be in the actual experimental data.  The next step would be to
add 'scan_args' to the run start too.

### DIFF
--- a/metadatastore/examples/sample_data/common.py
+++ b/metadatastore/examples/sample_data/common.py
@@ -80,7 +80,8 @@ def example(func):
         if run_start_uid is None:
             run_start_uid = insert_run_start(time=get_time(), scan_id=1,
                                            beamline_id='example',
-                                           uid=str(uuid.uuid4()))
+                                           uid=str(uuid.uuid4()),
+                                           scan_type=func.__name__)
 
         # these events are already the sanitized version, not raw mongo objects
         events = func(run_start_uid, sleep)
--- a/metadatastore/examples/sample_data/common.py
+++ b/metadatastore/examples/sample_data/common.py
@@ -78,10 +78,13 @@ def example(func):
     @wraps(func)
     def mock_run_start(run_start_uid=None, sleep=0, make_run_stop=True):
         if run_start_uid is None:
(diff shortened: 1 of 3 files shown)


tacaswell:
travis is producing a strange eror...

ericdill:
I think this is a pymongo version thing. travis is installing pymongo-3.0.3

ericdill:
This will get rebased on top of #176 once that PR is done

danielballan:
@ericdill What about this:
1. Delete all the sample data generation code from databroker.
2. Make bluesky (but not ophyd) a test dependency.
3. Generate data like:

``` python
from bluesky.tests.utils import setup_test_run_engine
from bluesky.register_mds import register_mds
RE = setup_test_run_engine()
register_mds(RE)

from bluesky.plans import AbsScanPlan
from bluesky.examples import motor, det
plan = AbsScanPlan([det], motor, 1, 5, 5)
RE(plan)
```

This didn't make sense originally because, at the time we were writing broker, there was no way to generate fake data with a simulated EPICS IOC. But now, examples like the above look a _lot_ simpler and more maintainable than the rats nest of manual document-generation code we have in place.

ericdill:
I am not sure where this makes the most sense.  I'm not a huge fan of picking up a bluesky dependency in metadatastore. That being said, the code snippet above is much nicer than the pile of code we currently have in databroker/metadatastore.  I am :+1: on _temporarily_ adding a bluesky dep to the sample data generation in mds/databroker with the ultimate goal of having a vertical integration repo actually working in the not to distant future.  At that point we can move the sample data generation code into this magical vaporware vertical integration repo

danielballan:
Sounds fine to me.

ericdill:
Alright great. So I'll close this PR.  Do you want to set up data generation code with bluesky?

danielballan:
I would like to see that happen, but I don't think I will do it personally any time soon.

ericdill:
Alright. I'll add it as an issue on both repos

PR #181, opened 2015-09-30
Pin pymongo to 2.6.3 on travis

ericdill:
if travis passes, this PR is finished


ericdill:
Ah fun... This error:

```
nose.proxy.TypeError: 'Collection' object is not callable. If you meant to call the 'initialize_ordered_bulk_op' method on a 'Collection' object it is failing because no such method 
```

is coming from pymongo 2.6.3 since `initialize_ordered_bulk_op` was introduced in 2.7.

metadatastore tests fail on pymongo > 3.0.0 and now fail on pymongo < 2.7.0.

ericdill:
Taking what I've learned here and applying it to #176

PR #184, opened 2015-10-16
Travis config is generated from a template

ericdill:
With this PR, metadatastore is using an autogenerated travis template.
The template can be found at https://github.com/ericdill/travis-little-helper. 
Instructions for creating a new travis yml can be found in the readme.

Also added some badges for current dev build on anaconda.org/lightsource2-dev/metadatastore 
and latest point release on anaconda.org/lightsource2/metadatastore


ericdill:
This is basically just a bunch of travis jiggery-pokery.  If anyone is interested in what is going on behind the scenes, I'm happy to explain. Otherwise, merge away. Or I can squash commits.  Let me know!

cc @danielballan @klauer @arkilic

conda-recipe/run_test.sh
```diff
@@ -0,0 +1 @@
+nosetests -v metadatastore --cover-package metadatastore --with-coverage
```
tacaswell, line 1: Does this get run on every build and does this require a running mongo instance?

README.md
```diff
@@ -2,10 +2,25 @@
...
+
 
 # metadatastore
 NSLS2 Beamlines metadatastore prototype implemented in MongoDB.
 
+## Update travis-ci configuration yaml
+
+```
+git clone https://github.com/ericdill/travis-little-helper
+cd travis-little-helpe
```
tacaswell, line 17: missing an 'r'

.travis.yml
```diff
@@ -1,51 +1,114 @@
...
+  - export CONDA_BUILD_COMMAND="conda build conda-recipe --python=$TRAVIS_PYTHON_VERSION"
+  - export GIT_FULL_HASH=`git rev-parse HEAD`
+  - export CLEAN_ANACONDA_CHANNEL=False
+  # SOME DEBUG OUTPUT
+  - echo "OWNER_NAME=$OWNER_NAME"
+  - echo "REPO_NAME=$REPO_NAME"
+  - echo "TRAVIS_COMMIT=$TRAVIS_COMMIT"
+  - git describe
+  # INSTALL DEPS FOR BUILDING AND UPLOADING
+  - conda install conda-build jinja2 anaconda-client
```
tacaswell, line 73: `jinja2` is _now_ a dep of `conda-build`, but leave it to be safe

conda-recipe/meta.yaml
```diff
@@ -1,56 +1,60 @@
...
+    - pytz
+    - pyyaml
+    - six
 
 test:
   requires:
-    - nslsii_dev_configuration
+    - boltons
+    - coverage
+    - mongoengine
```
danielballan, line 35: Don't we need to pin this?

conda-recipe/meta.yaml
```diff
@@ -1,56 +1,60 @@
...
-    - mongoengine ==0.8.7
-    - pymongo ==2.9
-    - six
-    - pyyaml
-    - prettytable
-    - humanize
-    - numpy
-    - pytz
-    - jinja2
     - boltons
```
danielballan, line 24: I think the only build deps are python and six.

.travis.yml
ericdill, line 73: yeah i saw that PR finally get merged. I agree that this should be left for now until we are sure that the most recent version of conda-build will install jinja2

conda-recipe/meta.yaml
ericdill, line 24: Unless I misunderstand, build dependencies are only those things that need to be installed for setup.py to run. Since the only imports in setup.py are setuptools/distutils and versioneer, I am confident that we do not need six.  However, as I type this I am not sure if we need setuptools/distutils or not

conda-recipe/meta.yaml
ericdill, line 35: Probably.  Also I recently discovered that test requires is a superset of run requires, so much of this section can just be deleted

conda-recipe/run_test.sh
ericdill, line 1: Yes and I forget what we decided on when the automated tests should be run during conda-build

danielballan, line 1: I think we decided against it.

ericdill, line 1: Tom and I had a conversation more recently, but I don't remember what exactly the outcome was

ericdill, line 1: Oh this is what is was:

We will test the conda build on Travis. We will test the conda build in the git repo. We will not include the test script with the recipes on conda-prescriptions. The logic being that our tagged versions have been thoroughly tested, but if you are building a conda package from a non tagged version it is better to test it during the build.

Also it is very easy to skip the testing by simply deleting the run_tests.sh file

danielballan, line 1: By "test conda build" you mean "run the full unit tests suite, not just
check a few imports"? Even on Travis where those tests were just run?

ericdill, line 1: Yes, i mean run the full test suite.

> Even on Travis where those tests were just run?

Yes. Those tests were run on the git repository, not the built conda package.  History has shown that it is quite simple for those two to _not_ be equivalent.

conda-recipe/meta.yaml
danielballan, line 24: Yes, setuptools too: https://github.com/NSLS-II/metadatastore/blob/master/setup.py#L2

I'd leave six there as a futureproof/convenience. Doesn't hurt.

conda-recipe/run_test.sh
danielballan, line 1: Sorry if I seem to be pushing back on this too hard, but I've never seen another project do this and debugging Travis can eat a lot of time as we know. What particular situation are you trying to catch? Where the `.travis.yml` and the `meta.yml` are out of sync, with respect to versions of deps?

ericdill, line 1: No, the problem I am trying to avoid is when `python setup.py install` and `conda install` produce different results

tacaswell:
Is this still a going concern?  If so it needs a rebase.

PR #185, opened 2015-10-25
Use the correct license

ericdill:


tacaswell:
wait, what?

Also, leave the year an `2014-` so we don't have to update these every year.

ericdill:
I don't understand your "wait what?"

tacaswell:
How did we end up with a different license to begin with?

tacaswell:
The current license is the EPICS license: http://www.aps.anl.gov/epics/license/open.php 

:+1: on moving to a more well known license

danielballan:
I'll let @arkilic do the honors.

ghost:
Goodbye epics license

PR #189, opened 2016-01-18
Api match mds schema

tacaswell:


tacaswell:
Tests pass, but we do not have a doct conda package it can see yet.

.travis.yml
```diff
@@ -24,7 +24,7 @@ install:
 - conda config --add channels lightsource2
 - conda create -n testenv pip nose python=$TRAVIS_PYTHON_VERSION pymongo six pyyaml numpy pandas jinja2 boltons prettytable humanize
 - source activate testenv
-- pip install mongoengine
+- 'pip install https://github.com/NSLS-II/doct/zipball/master#egg=doct'
```
ericdill: You can `conda install doct` now. I built master and uploaded it to `lightsource2`

tacaswell:
@arkilic @danielballan In metadataservice the events are stored with data/timestamp as two dictionaries.  In this implementation (and previously) they are stored as a `(value, timestamp)` tuple in a single `data` dict.

#### Pros of changing schema (now)
- already doing a migration, so this is a great time to do it if we are going to do it
- store on disk matching the in-memory representation
- reduces thrashing on both way in and out

##### cons of changing schema
- store in form that matches `read`, however that data gets re-formatted on the way out of bs to the `{data:{}, timestamp: {}}` form.  
- current schema help to enforce that we _always_ have value + timestamp.

On net, I think changing the schema now is the better option.

ghost:
Agreed. Less migrations :+1:

PR #192, opened 2016-02-05
Fix search order

tacaswell:


ghost:
Did you check how much overhead sorting on the cursor adds, especially for large number of events?

tacaswell:
Part of the API is that the events come out in the order they went in.  We are indexing on time so if I am reading the docs right it should not be a big deal.

danielballan:
Right. Removing sorting amounts to an API break, and indeed databroker tests are failing because of this change. @arkilic, if we want to get a performance boost from dropping sorting, that can be considered -- with coordinated changes downstream.

ghost:
I checked this out with test data, there is no issue in terms of performance.

PR #194, opened 2016-02-12
FIX: Ensure port is an int.

danielballan:
When defining the `MDS_PORT` via a environmental variable, I get

``` python
TypeError: port must be an instance of int
```

This resolves the issue at the lowest level.

### DIFF
--- a/metadatastore/conf.py
+++ b/metadatastore/conf.py
@@ -62,6 +62,9 @@ def load_configuration(name, prefix, fields):
     for field in fields:
         var_name = prefix + '_' + field.upper().replace(' ', '_')
         config[field] = os.environ.get(var_name, config.get(field, None))
+        # Valid values for 'port' are None or castable to int.
+        if field == 'port' and config[field] is not None:
+            config[field] = int(config[field])
 
     missing = [k for k, v in config.items() if v is None]
     if missing:


metadatastore/commands.py
```diff
@@ -65,7 +65,7 @@ def reconfigure(self, config):
     def _connection(self):
         if self.__conn is None:
             self.__conn = MongoClient(self.config['host'],
-                                      self.config.get('port', None))
+                                      int(self.config.get('port', None)))
```
ericdill: `int(None)` will raise.  `None` is a valid option for the port in `MongoClient.__init__`.  Therefore I think this code needs to be:

``` python
if self.__conn is None:
    port = self.config.get('port')
    if port:
        port = int(port)
    self.__conn = MongoClient(self.config['host'], port)
```

or

``` python
 int(self.config.get('port', 27017)))
```

ericdill:
Good catch. That has bitten me previously too.  I wonder what regression caused it to surface again

metadatastore/commands.py
danielballan: @arkilic Any opinion on this?

ericdill: @arkilic bump.

tacaswell: should probably do this casting in conf, not here.

ghost: I agree with Tom, don't let this leak out of conf because I use the same conf in a few other places and I certainly forget to cast it at times

I like: 

```
int(self.config.get('port', 27017)))
```

less code lol

danielballan:
The [code that originally fixed this](https://github.com/NSLS-II/metadatastore/pull/153) has been refactored away. With help from @tacaswell I fixed it at the source. Please please someone merge so I can stop seeing this error.

ericdill:
LGTM. Sorry this sat for so long @danielballan.

PR #195, opened 2016-02-16
Enh listify ndarray

tacaswell:


ericdill:
Where is `_transform_data` even used? I can't find it anywhere in mds/bs/ophyd

metadatastore/commands.py
```diff
@@ -839,6 +841,11 @@ def _transform_data(data, timestamps):
     to storage format:
         {'data': {<key>: (<value>, <timestamp>)}.
     """
+    # shallow copy so we can do some norrmalization
```
ericdill: `s/norrmalization/normalization`

metadatastore/commands.py
```diff
@@ -839,6 +841,11 @@ def _transform_data(data, timestamps):
...
+    data = dict(data)
```
ericdill, line 845: isn't @stuwilkins doing something where the data and timestamps are both numpy arrays?  Does it makes sense to do the same thing for `timestamps` as you are doing for `data`?

tacaswell:
```
11:40 $ git grep _transform_data
metadatastore/commands.py:    val_ts_tuple = _transform_data(data, timestamps)
metadatastore/commands.py:            val_ts_tuple = _transform_data(ev['data'], ev['timestamps'])
metadatastore/commands.py:def _transform_data(data, timestamps):
```

metadatastore/commands.py
tacaswell, line 845: I do not think so as without this PR if you try to put an np.ndarray into mongo it dies because it does not know how to encode the ndarray into json.   

I do not think we want to leave the model where a timestamp is a single scalar.  If you want to put a time series into an event and the exact timestamps are important then that should go in as a data key.

ericdill:
Ah, I was on the master branch.

metadatastore/commands.py
ericdill, line 845: Oh, is that what @stuwilkins is doing?  Putting the timestamps data in as a data key?

ericdill:
LGTM. Anything else you want to add?

tacaswell:
This is a hot-fix to make the single-element MCA at SRX running (and ken just confirmed that it works for this use-case).

PR #196, opened 2016-02-16
WIP: Enh global state

tacaswell:


tacaswell:
@arkilic @danielballan This code base should be able to:
- write documents compatible with the new mdservice schema (uid foreign keys, data + timestamp dictionaries)
- read both old (mongoengine produced documents, oid foreign keys) and new sechema
- using the classes in `metadatastore.mds` connect to multiple mongo instances in the same process

I suggest the following course of action tomorrow
- use this branch + metadataclient to verify reading header (start, stop, descriptor) documents between the old db and the transfer to mdservice that Arman has already done.  This should make you confident that this code can correctly deal with the old schema
- use this branch to compare header information between the old database and the metadataclient database (to convince you this can read the new schema)
- use this branch do to spot checks of get_events reading from the old database and another process running the v0.3.x branch.
- do the dump -> reload process into the new database
- use this branch to compare header + event data between the new and old databases

tacaswell:
It is probably worth looking into basing metadataservice on the `core` functions in this library to pick up the caching. 

Documentation is still on the to-do list.

ghost:
Any more love for this pr @tacaswell ?

tacaswell:
Yes, the last commit broke things, I want to un-break things.  I think having a `core_v0` and a `core` (`core_v1`) is a better way to do this (instead of adding rouge version numbers to the calls).

ghost:
Great, let me know when this is good to go over

tacaswell:
I think this is ready to review.

tacaswell:
Not sure if it is worth sorting out a way to test the v0 code.  I think once we get all of the beamlines migrated (hopefullly this shutdown) we can just delete it.

ghost:
We have performed the suggested tests above and were able to retrieve data from the migrated service database (that has the latest foreign key schema) and legacy metadatastore.

PR #201, opened 2016-03-23
Update conf.py

CJ-Wright:
I think this might be a bit more clear, also why is there a trailing `_` on the env variable?


danielballan:
Agreed. Thanks, @CJ-Wright.

We have a special conda package, `conda-etc`, that defines `CONDA_ETC_`, the path to `/etc` in the conda env. We used the trailing slash in case upstream conda ever decided to use `CONDA_ETC` themselves.

All of this machinery is a little too clever and is probably headed for retirement.

PR #203, opened 2016-03-28
ENH: Switching to pytest

ericdill:


ericdill:
Ready for review. attn @tacaswell @klauer

metadatastore/test/test_sample_data.py
```diff
@@ -1,42 +1,39 @@
...
 
-def test_noisy_for_smoke():
-    vals = [0, np.arange(0, 5)]
-    for v in vals:
-        yield _noisy_helper, v
+@pytest.mark.parametrize(
+    'failing_time, mod',
+    itertools.product([-1, 1], [temperature_ramp, multisource_event])
+)
+def test_sleepy_failures(failing_time, mod):
```
tacaswell, line 33: What is this testing and can we just remove that code path?

ericdill, line 33: It is testing to make sure that our data generation code, in this case `temperature_ramp` and `multisource_event` will correctly fail if they are given a sleep time.  I do not care either way if we keep or remove this test.

ericdill:
Note the move of `metadatastore.utils.testing.py` -> `metadatastore.test.utils.py`. Reasoning is that we dont need a utils subpackage for a testing thing. Just move it into the test subpackage and get rid of the now unnecessary utils subpackage

metadatastore/test/test_sample_data.py
```diff
@@ -1,42 +1,39 @@
...
-from metadatastore.utils.testing import mds_setup, mds_teardown
+from .utils import mds_setup, mds_teardown
 import numpy as np
 import itertools
-from nose.tools import raises
+import pytest
 
-def setup():
+
+def setup_function(function):
```
tacaswell, line 10: do we really want to do this on a per-function not per-module basis?  Saw playing with data broker that this majorly slows down the tests.

ericdill, line 10: test suite takes 1.36 seconds to run on my machine. Is that too slow?

tacaswell, line 10: it was taking seconds _per test_ with databroker, but if this is going fast enough, then it is fine.

ericdill, line 10: test suite takes 0.67 seconds if `s/function/module`

PR #208, opened 2016-03-30
MNT: Sort events so oldest is first

ericdill:
Fixes the problem where events are returned newest first. This causes issues in suitcase.

https://gist.github.com/ericdill/11c77848fee4725efdef2a9d4e2fc2c6

### DIFF
--- a/metadatastore/core.py
+++ b/metadatastore/core.py
@@ -1007,7 +1007,7 @@ def find_events(start_col, start_cache,
 
     _format_time(kwargs, tz)
     col = event_col
-    events = col.find(kwargs)
+    events = col.find(kwargs, sort=[('time', pymongo.ASCENDING)])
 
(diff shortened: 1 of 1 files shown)


ericdill:
Ok so my solution in NSLS-II/suitcase#8 was to just use metadatastore.get_events_generator. This can wait.

tacaswell:
Did you mean to close this?

ericdill:
Yes. @arkilic had some reservations about sorting the events differently in `get_events()` which is fine by me as I don't need to use `get_events()` to achieve my objectives for NSLS-II/suitcase#8.

tacaswell:
We are using an identical sort on `get_events_generator`

ericdill:
@tacaswell I really don't care what we do here. Do you want this reopened so you can merge it?

PR #211, opened 2016-04-26
ENH: Add generic insert dispatcher.

danielballan:
This obviates `bluesky/register_mds.py`. Registering metadatastore becomes as simple as:

``` python
RE.subscribe('all', metadatastore.commands.insert)
```

### DIFF
--- a/metadatastore/mds.py
+++ b/metadatastore/mds.py
@@ -669,12 +669,12 @@ class MDS(MDSRO):
                                        events=events,
                                        validate=validate)
 
-    def insert(self, name, doc, validate=False):
+    def insert(self, name, doc):
         if name != 'bulk_events':
-            getattr(self, self._INS_METHODS[name])(validate=validate, **doc)
+            getattr(self, self._INS_METHODS[name])(**doc)
         else:
             for desc_uid, events in doc.items():
                 # If events is empty, mongo chokes.
                 if not events:
                     continue
-                self.bulk_insert_events(desc_uid, events, validate=validate)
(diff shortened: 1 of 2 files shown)


danielballan:
cc @arkilic I see some problems with the implementation in metadataclient (it was [this commit](https://github.com/arkilic/metadataclient/commit/aaa9da6245267b8498eb92870ce2b7d13a8f9419)). Take a look at this code.

ghost:
@danielballan I remember looking at this together. What is the problem here?
I think we could look at the `name == bulk_insert_events` instead of dtype

metadatastore/mds.py
```diff
@@ -501,6 +501,15 @@ def find_last(self, num=1):
 
 
 class MDS(MDSRO):
+    def __init__(self, *args, **kwargs):
+        self.__doc__ = super().__init__.__doc__
+        super().__init__(*args, **kwargs)
+        self._INS_METHODS = {'start': self.insert_run_start,
```
tacaswell: due to the way that method look up works, this might be better to map to strings and then use `getattr` in `insert`.

danielballan:
I caught a couple problems:
- spelling: 'bulk_events' not 'bulk_event'; 'start' not 'run_start'; 'stop' not 'run_stop'
- bulk events data type: it's dict, not a list -- see my implementation
- bulk events edge case: short-circuit if there are no events

ghost:
The service layer takes care of these edge cases in terms of validation etc.
I will take care of the others.

PR #215, opened 2016-04-28
ENH: Add a wildcard text index

danielballan:
As we migrate to Mongo 3.2 (planned for tomorrow) we can support fast text search, which I demonstrated awhile ago in [this toy project, 'broker search'](https://github.com/NSLS-II/broker-search)

[Relevant section of Mongo docs](https://docs.mongodb.org/manual/core/index-text/#wildcard-text-indexes)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,12 @@ language: python
 sudo: false
 services:
 - mongodb
+addons:
+  apt:
+    sources:
+    - mongodb-3.2-precise
+    packages:
+    - mongodb-org-server
 matrix:
   include:
   - python: 2.7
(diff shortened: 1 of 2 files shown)


tacaswell:
:+1: leave to @arkilic to merge.  Can we support the same thing in client?

licode:
Nice feature, so we can perform search on all fields easily. Do we need to include this also on event or descriptor?

tacaswell:
Definitely not on events, maybe on descriptors.  That would solve the "I know the motor was called 'ssx'" problem

licode:
The search on events will be too heavy. And all important information are actually stored in start/stop/descriptors.

danielballan:
This PR already includes an index on descriptor. Only event is left out.

licode:
👍

PR #218, opened 2016-04-29
work around LRU bug for now

danielballan:
simply replaces fancy (but buggy) LRU cache with a dict, the simplest cache of them all


danielballan:
(The plan here is to reinstate the LRU cache once @tacaswell has the bandwidth to hunt down a hard-to-reproduce bug in the cache.)

PR #219, opened 2016-04-29
Add time indexes for descriptor and event

danielballan:
Mongo raises an error on `find_descriptors()` because, although we have a joint index of run_start and time, we don't have one just one time. This seems like a safe and useful thing to add.

Also, the migration verification script revealed we were _still_ missing some sorts in core_v0, which is here remedied.

### DIFF
--- a/metadatastore/core.py
+++ b/metadatastore/core.py
@@ -357,7 +357,8 @@ def get_events_generator(descriptor, event_col, descriptor_col,
                                       run_start_cache)
     col = event_col
     ev_cur = col.find({'descriptor': descriptor_uid},
-                      sort=[('time', pymongo.ASCENDING)])
+                      sort=[('descriptor', pymongo.DESCENDING),
+                            ('time', pymongo.ASCENDING)])
 
     data_keys = descriptor['data_keys']
     for ev in ev_cur:
@@ -1019,7 +1020,9 @@ def find_events(start_col, start_cache,
 
     _format_time(kwargs, tz)
     col = event_col
-    events = col.find(kwargs)
+    events = col.find(kwargs,
+                      sort=[('descriptor', pymongo.DESCENDING),
+                            ('time', pymongo.ASCENDING)])
 
     for ev in events:
         ev.pop('_id', None)
--- a/metadatastore/core_v0.py
+++ b/metadatastore/core_v0.py
@@ -369,7 +369,8 @@ def get_events_generator(descriptor, event_col, descriptor_col,
     col = event_col
     oid = _UID_TO_OID_MAP[descriptor_uid]
     ev_cur = col.find({'descriptor_id': oid},
-                      sort=[('time', pymongo.ASCENDING)])
+                      sort=[('descriptor_id', pymongo.DESCENDING),
+                            ('time', pymongo.ASCENDING)])
 
     data_keys = descriptor['data_keys']
     for ev in ev_cur:
@@ -986,7 +987,8 @@ def find_descriptors(start_col, start_cache,
     _format_time(kwargs, tz)
 
     col = descriptor_col
-    event_descriptor_objects = col.find(kwargs)
+    event_descriptor_objects = col.find(kwargs,
+                                        sort=[('time', pymongo.ASCENDING)])
 
     for event_descriptor in event_descriptor_objects:
         yield _cache_descriptor(event_descriptor, descriptor_cache,
@@ -1035,7 +1037,9 @@ def find_events(start_col, start_cache,
 
     _format_time(kwargs, tz)
     col = event_col
-    events = col.find(kwargs)
+    events = col.find(kwargs,
+                      sort=[('descriptor_id', pymongo.DESCENDING),
+                            ('time', pymongo.ASCENDING)])
 
     for ev in events:
         ev.pop('_id', None)
--- a/metadatastore/mds.py
+++ b/metadatastore/mds.py
@@ -119,14 +119,24 @@ class MDSRO(object):
     @property
     def _descriptor_col(self):
         if self.__descriptor_col is None:
+            # The name of the reference to the run start changed from
+            # 'run_start_id' in v0 to 'run_start' in v1.
+            if self.version == 1:
+                rs_name = 'run_start'
+            elif self.version == 0:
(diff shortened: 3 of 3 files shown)


danielballan:
OK, after quick discussion with @arkilic, this PR has been changed. No new indexes are being added to event. The important thing is that all the sorting imposed by our built-in queries has a matching index.

metadatastore/core.py
```diff
@@ -357,7 +357,8 @@ def get_events_generator(descriptor, event_col, descriptor_col,
                                       run_start_cache)
     col = event_col
     ev_cur = col.find({'descriptor': descriptor_uid},
-                      sort=[('time', pymongo.ASCENDING)])
```
danielballan, line 360: i.e., Don't sort by time because there is no index for time.

metadatastore/core.py
```diff
@@ -1019,7 +1020,9 @@ def find_events(start_col, start_cache,
 
     _format_time(kwargs, tz)
     col = event_col
-    events = col.find(kwargs)
+    events = col.find(kwargs,
```
danielballan, line 1023: This was mistakenly not being sorted at all.

metadatastore/core_v0.py
```diff
@@ -369,7 +369,8 @@ def get_events_generator(descriptor, event_col, descriptor_col,
     col = event_col
     oid = _UID_TO_OID_MAP[descriptor_uid]
     ev_cur = col.find({'descriptor_id': oid},
-                      sort=[('time', pymongo.ASCENDING)])
```
danielballan, line 372: Again, don't sort on time because there is no index on time.

metadatastore/core_v0.py
```diff
@@ -986,7 +987,8 @@ def find_descriptors(start_col, start_cache,
     _format_time(kwargs, tz)
 
     col = descriptor_col
-    event_descriptor_objects = col.find(kwargs)
+    event_descriptor_objects = col.find(kwargs,
```
danielballan, line 990: This PR adds a 'time' index for descriptors, so these can be returned in a deterministic order, which helps migration validation.

metadatastore/core_v0.py
```diff
@@ -1035,7 +1037,9 @@ def find_events(start_col, start_cache,
 
     _format_time(kwargs, tz)
     col = event_col
-    events = col.find(kwargs)
+    events = col.find(kwargs,
```
danielballan, line 1040: Ditto.

metadatastore/mds.py
```diff
@@ -106,10 +108,11 @@ def _descriptor_col(self):
 
             self.__descriptor_col.create_index([('uid', pymongo.DESCENDING)],
                                                unique=True)
-            self.__descriptor_col.create_index(
-                [('run_start', pymongo.DESCENDING),
-                 ('time', pymongo.DESCENDING)],
-                unique=False, background=True)
+            self.__descriptor_col.create_index([('run_start', pymongo.DESCENDING),
```
danielballan: (This is just reformatted.)

metadatastore/core_v0.py
tacaswell, line 372: These documents do not have a `descriptor` field.  Mongo will use the compound index here automatically ( https://docs.mongodb.org/manual/tutorial/sort-results-with-indexes/#sort-and-index-prefix ).  Once you have the index added, sorting on just time here is ok.

metadatastore/core_v0.py
```diff
@@ -1035,7 +1037,9 @@ def find_events(start_col, start_cache,
...
+                      sort=[('descriptor', pymongo.DESCENDING),
```
tacaswell: needs to be `'descriptor_id'`

metadatastore/mds.py
```diff
@@ -78,6 +78,8 @@ def _runstart_col(self):
 
             self.__runstart_col.create_index([('uid', pymongo.DESCENDING)],
                                              unique=True)
+            self.__runstart_col.create_index([('time', pymongo.DESCENDING)],
```
tacaswell: We do not need this index, we have other index that starts with time https://docs.mongodb.org/manual/core/index-compound/#prefixes

metadatastore/mds.py
tacaswell: This needs to have the v0 vs v1 distinction added.

danielballan: Why?

metadatastore/mds.py
danielballan: Agreed.

metadatastore/core_v0.py
danielballan: Done.

metadatastore/core_v0.py
danielballan, line 372: We aren't adding a time index on events (for insert perf reasons) but I will change `descriptor` -> `descriptor_id`.

metadatastore/mds.py
tacaswell: because in v0 it in `run_start_id` and in v1 it is `run_start`

metadatastore/core.py
```diff
@@ -357,7 +357,8 @@ def get_events_generator(descriptor, event_col, descriptor_col,
                                       run_start_cache)
     col = event_col
     ev_cur = col.find({'descriptor': descriptor_uid},
-                      sort=[('time', pymongo.ASCENDING)])
+                      sort=[('descriptor', pymongo.DESCENDING),
```
tacaswell, line 360: ~~I do not think this should have the `'descriptor'` in the sort as now the time is essentially not used.~~  But there is only one descriptor here anyway.

metadatastore/core_v0.py
```diff
@@ -369,7 +369,8 @@ def get_events_generator(descriptor, event_col, descriptor_col,
     col = event_col
     oid = _UID_TO_OID_MAP[descriptor_uid]
     ev_cur = col.find({'descriptor_id': oid},
-                      sort=[('time', pymongo.ASCENDING)])
+                      sort=[('descriptor_id', pymongo.DESCENDING),
```
tacaswell, line 372: same concern about stacking the keys here.

PR #220, opened 2016-04-29
ENH: be able to pickle MDS objects

tacaswell:


metadatastore/mds.py
```diff
@@ -31,6 +31,25 @@ def __init__(self, config, version=1):
         self._api = None
         self.version = version
 
+    def __getstate__(self):
+        return self.version, self.config
+
+    def __setstate__(self, state):
+        self._RUNSTART_CACHE = {}
```
klauer, line 44: Can all this go in some sort of `_reset_state` or something?

danielballan, line 44: +1

danielballan, line 44: Actually, it might be better to do reset_state separately:

``` python
self._RUNSTART_CACHE.clear()
...
```

tacaswell, line 44: When objects are unpickled their `__init__` is not run.

danielballan, line 44: (done, in a PR against this PR)

metadatastore/mds.py
```diff
@@ -17,9 +17,17 @@ def __init__(self, config, version=1):
         self._RUNSTART_CACHE = {}
         self._RUNSTOP_CACHE = {}
         self._DESCRIPTOR_CACHE = {}
-
+        self.reset_connection()
```
tacaswell, line 20: I am not a huge fan of this because now it is hard to tell what state is in this object but eh

metadatastore/mds.py
licode, line 44: good to know how to implement default methods for pickle. +1

PR #222, opened 2016-05-04
Enh v0 insert

tacaswell:


tacaswell:
I do not understand the warning failure only on py2.7.

Both versions go through identical code paths up to where the warning should be raised.

@danielballan do you know if there are changes between how warnings are suppressed between 2 and 3?

danielballan:
Hmm. That's new to me.

danielballan:
What happens if you run `run_tests.py` with `-W`?

danielballan:
Very nice test pattern. I would like to review this with fresh eyes tomorrow.

metadatastore/mds.py
```diff
@@ -106,9 +106,13 @@ def _runstart_col(self):
     def _runstop_col(self):
         if self.__runstop_col is None:
             self.__runstop_col = self._db.get_collection('run_stop')
-
-            self.__runstop_col.create_index([('run_start', pymongo.DESCENDING),
-                                            ('uid', pymongo.DESCENDING)],
+            if self.version == 0:
+                self.__runstop_col.create_index('run_start_id',
+                                                unique=True)
+            else:
```
danielballan, line 112: When I used this pattern in the `_events_col`, you suggested that the `else` block should raise. I actually disagree; I think it's fine to assume that if the code is unchanged, v2 can assume it uses the same indices as v1. So as far as I'm concerned this is fine, but Past Tom objects. :- D

danielballan:
Thanks for doing this. Consider my one line comment and then merge at will.

metadatastore/test/test_mds.py
```diff
@@ -1,11 +1,401 @@
...
+                                         beamline_id='testbed',
+                                         owner='tester',
+                                         group='awesome-devs',
+                                         project='Nikea',
+                                         time=document_insertion_time,
+                                         uid=str(uuid.uuid4()))
+    time = ttime.time()
+    # test insert
+    ev_desc_uid = mds.insert_descriptor(run_start_uid, data_keys, time,
+                                        str(uuid.uuid4()))
```
licode, line 97: There is a time argument for both insert_run_start, and insert_run_stop. So we can get the total time for this run. Do we need time for descriptor, or maybe this time should be the same as run_start time?

metadatastore/test/test_mds.py
```diff
@@ -1,11 +1,401 @@
...
+        run_start_uid = mds.insert_run_start(
+            scan_id=30220, beamline_id='testbed',
+            owner='Al the Aardvark', group='Orycteropus',
+            project='Nikea', time=ttime.time(),
+            uid=run_start_uid, custom={'order': 'Tubulidentata'})
+
+    rs = next(mds.find_run_starts(order='Tubulidentata'))
+    assert rs['uid'] == run_start_uid
+
+    with warnings.catch_warnings(record=True) as w:
```
licode, line 126: What kind of warnings do we usually receive?

metadatastore/test/test_mds.py
```diff
@@ -1,11 +1,401 @@
...
+                 'data_key3': {'source': 'PV:pv1',
+                               'shape': [],
+                               'dtype': 'number',
+                               'external': 'FS:foobar'}}
+
+    document_insertion_time = ttime.time()
+    run_start_uid = mds.insert_run_start(scan_id=3022013,
+                                         beamline_id='testbed',
+                                         owner='tester',
+                                         group='awesome-devs',
```
licode, line 90: I like the group name.
When reading some databroker information, I find owner/group information is probably not used correctly. The owner is the beamline scientist, and group is the user group? Sorry for the tedious questions.

danielballan, line 90: There is no established standard. Ultimately, this kind of info should match up with the information in a beam time proposal. There will be a standard some day. But for now it's up to users and beamline scientists to enter useful, accurate info. We have no way of verifying it at this time.

metadatastore/test/test_mds.py
danielballan, line 97: Every document had a time. The times on the start, stop, and descriptor documents aren't necessarily scientifically important or precise -- they just tell us wen the document was inserted. The timestamps field in an event document (and, to some extent, the time field in an event document) are more scientifically meaningful.

metadatastore/test/test_mds.py
danielballan, line 126: In this example, the keyword 'custom' is used,  and that is an old feature now deprecated. (It is faster and simpler if custom fields are entered as extra keyword arguments instead of being nested inside a dictionary the way they are here.) When a user passes extra metadata through 'custom' it still works but it raises a warning recommending passing keyword arguments instead.

danielballan, line 126: P.S. No problem! It's good to have eyes on this and more people who come to understand the details.

metadatastore/mds.py
tacaswell, line 112: I will ignore past Tom in this case.

metadatastore/test/test_mds.py
licode, line 90: Got it. It is easy to search in terms of users/group if we can put them in a standard way later.

metadatastore/test/test_mds.py
licode, line 126: Agree. It should be faster if we don't put custom as a nested dict. thanks!

ghost:
@danielballan @tacaswell Does this conclude the indexing work?

tacaswell:
@arkilic Hopefully, but we will know once we get it into the wild.

ghost:
@tacaswell Cool, I will start aligning the mdc once we feel good about it. 
For those that do not have the context, I am not concerned because the indexing discussed here as it deals with query result cursors, not inserts. Nothing critical that would require rebuilding the db indexes or migration.

tacaswell:
I would also add that these indecies are things that can (should?) be tuned on a per-beamline basis.

ghost:
Yes, I came across a tool that monitors the most common queries and allows us to optimize the application accordingly. So far it looks pretty nifty but pricy give the amount of data we store in mongo ($2.50 per GB/month)

metadatastore/test/test_mds.py
cowanml, line 90: On Sat, 7 May 2016, Dan Allan wrote:
...

> On Sat, 7 May 2016, Li Li wrote:
> 
> > -    run_start_uid = mds.insert_run_start(scan_id=3022013,
> > -                                         beamline_id='testbed',
> > -                                         owner='tester',
> > -                                         group='awesome-devs',
> >   ...
> >   When reading some databroker information, I find owner/group
> >   information is probably not used correctly. The owner is the
> >   beamline scientist, and group is the user group? Sorry for the
> >   tedious questions.
> 
> There is no established standard. Ultimately, this kind of info
> should match up with the information in a beam time proposal. There
> will be a standard some day. But for now it's up to users and
> beamline scientists to enter useful, accurate info. We have no way
> of verifying it at this time.

To make things more clear, we could change 'group' to 'proposal' and
'owner' to 'collector'.

A proposal is the standard unit of access control at nsls2 and most
facilities, so we need the proposal.  The pi of the proposal
ultimately owns the data, so with the proposal, the actual 'owner'
(the pi) would be redundant.

The pi is very often not the person collecting data though and it
makes sense to record who does.

We might not have a documented standard or connection to pass (yet),
but the basic needs and requirements are clear, not remotely new, and
there are examples at a bunch of facilities.

Until we are collecting data using individual accounts (either at the
os or application level) and have a link to pass, it will still be
up to users and beamline scientists to enter useful, accurate info,
but we do know what fields they should be entering.

-matt

danielballan, line 90: @cowanml Great. Would you write this up in a .rst file and submit a PR to NSLS-II/docs?

PR #225, opened 2016-05-18
API: commands api uses v1 schema

tacaswell:
attn @arkilic @ericdill 

Once this is merged we need a new tag (v0.6.0) ASAP.  This is so that mds does not try to insert v0 documents into the newly migrated v1 databases.


ghost:
Ready for tagging and deployment @tacaswell @ericdill

ghost:
@tacaswell Does this complete your verification in 5-id so we can roll it out to all beamlines?

tacaswell:
let me add one more commit to poison v0 / v1 crosses

PR #228, opened 2016-05-25
ENH:Ensure no _id leaks into docs as well

ghost:
🚲 🏠


metadatastore/test/test_mds.py
```diff
@@ -20,7 +20,7 @@ def check_for_id(document):
     """
     with pytest.raises(KeyError):
         document['id']
-
```
tacaswell, line 23: They need to be in thier own context managers, if the first one raises the second one will not be tested.

PR #231, opened 2016-05-31
MNT: Do not resolve configuration at import time.

danielballan:
This is what I had in mind. A conf dict with `None`s silently filled in (https://github.com/NSLS-II/metadatastore/pull/230) seems more potentially confusing than a KeyError on importing `metadatastore.conf`.

Would this change, which is analogous to how FS works, provide what metadataservice needs?


ghost:
Yes, it provides what mdservice needs

PR #232, opened 2016-05-31
Sentinels

danielballan:
As discussed offline, this future-proofs MDS for the next schema change (which we all hope won't happen for a long time). Deploying it will require running

``` python
from metadatastore.utils import install_sentinels
import metadatastore.conf
install_sentinels(metadatastore.conf.connection_config, 1)
```

just as we did for filestore v0.4.0 deployment.

This copies the sentinels-related code from FS verbatim with only two changes:
- The list of `versioned_collections` has been updated in `install_sentinels` and `self._db` (which checks for sentinels) 
- `create_test_database` returns a config with a timezone -- the local timezone from `tzlocal`

The MDS test utils have also been updated to use `create_test_database`.

### DIFF
--- a/metadatastore/examples/sample_data/common.py
+++ b/metadatastore/examples/sample_data/common.py
@@ -69,28 +69,3 @@ def noisy(val, sigma=0.01):
         return val + sigma * np.random.randn()
     else:
         return val + sigma * np.random.randn(len(val)).reshape(val.shape)
-
-
-get_time = ttime.time
-
-
-def example(func):
-    @wraps(func)
-    def mock_run_start(run_start_uid=None, sleep=0, make_run_stop=True):
-        if run_start_uid is None:
-            run_start_uid = insert_run_start(time=get_time(), scan_id=1,
-                                           beamline_id='example',
-                                           uid=str(uuid.uuid4()))
-
-        # these events are already the sanitized version, not raw mongo objects
-        events = func(run_start_uid, sleep)
-        # Infer the end run time from events, since all the times are
-        # simulated and not necessarily based on the current time.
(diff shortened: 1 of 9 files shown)


metadatastore/examples/sample_data/common.py
```diff
@@ -69,28 +69,3 @@ def noisy(val, sigma=0.01):
         return val + sigma * np.random.randn()
     else:
         return val + sigma * np.random.randn(len(val)).reshape(val.shape)
-
-
-get_time = ttime.time
-
-
-def example(func):
```
danielballan, line 77: This was overbuilt (by me) in the first place, and I'm removing it now because it makes it impossible to inject which MDS instance to use at runtime.

danielballan, line 77: (Now there is some code duplication in the individual examples, but there are only two examples and soon we want to replace the sample data generation with actual bluesky anyway.)

metadatastore/mds.py
```diff
@@ -86,6 +86,21 @@ def _db(self):
...
+                                        'event_descriptor', 'event']
+                for col_name in versioned_collection:
+                    val = sentinel.find_one({'collection': col_name})
+                    if val is None:
+                        raise RuntimeError('there is no version sentinel for '
+                                           'the {} collection'.format(col_name)
+                                           )
+                    if val['version'] != self.version:
+                        raise RuntimeError('DB version {!r} does not match'
+                                           'API version of FS {} for the '
```
tacaswell: fs -> mds

danielballan:
bump @tacaswell

danielballan:
To be continued in NSLS-II databroker

PR #233, opened 2016-06-01
Remove unused boltons imports

ghost:


tacaswell:
we should sort out how to re-impelment a LRU on this.

ghost:
@tacaswell we should either give [pylru ](https://pypi.python.org/pypi/pylru) or [joblib](https://pythonhosted.org/joblib/) a shot here. I think joblib would be excellent for filestore as caches numpy arrays like no other tool out there. Once we decide which library, I can add the caching layer.

tacaswell:
also, awj which naturally caches dataframes ;)

ghost:
Right, this feature is in awj. I think you can increase the performance with joblib. Let's discuss this in person. (also, we can make a decision on which lib to use)

PR #235, opened 2016-06-16
FIX: typo

danielballan:


metadatastore/mds.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, config, version=1):
         self.version = version
 
     def reset_caches(self):
-        self._RUNSTART_CAHCE.clear()
+        self._RUNSTART_CACHE.clear()
         self._RUNSTOP_CAHCE.clear()
```
klauer: This and next line too

danielballan: Bah, you _just_ beat me to it.

klauer: 😄

tacaswell:
`mds_all` includes the `commands` module as one of the categories,  I think there is a fixture which is just the two mds objects.

danielballan:
bump @arkilic @tacaswell This works now that commands is gone.

PR #239, opened 2016-07-29
ENH: Extend Event Document API to track external keys and filled-ness.

danielballan:
Presently, Events don't track whether any external data has been been "filled," and it's easy to write bugs that try to fill an Event twice, blowing up on a filestore error.

The obvious solution would be to stop filling Events in place, but that would mean copying _a lot_ of data in many common cases.

Instead, this PR introduces another parallel dict inside Event:

```
{'data': {'motor': 5, 'image': 'some_uid'},
 'timestamps', {'motor': ..., 'image': ...},
 'filled': {'image': False}
```

When `Broker.fill_event` operates on the document above, it will flip `False` to `True` when replacing `some_uid` with the image array data via filestore. Notice that fields that are not external (like 'motor' above) are not included in `filled`. Whether or not an Event has already been filled, the keys of the `filled` dict provide a quick way to determine which if any fields are external.

The new `filled` features has be introduced here in metadatastore, not databroker, because docts are immutable.

### DIFF
--- a/metadatastore/core.py
+++ b/metadatastore/core.py
@@ -362,6 +362,8 @@ def get_events_generator(descriptor, event_col, descriptor_col,
                             ('time', pymongo.ASCENDING)])
 
     data_keys = descriptor['data_keys']
+    external_keys = [k for k in data_keys if 'external' in data_keys[k]]
+    filled = {k: False for k in external_keys}
     for ev in ev_cur:
         # ditch the ObjectID
         del ev['_id']
@@ -374,6 +376,10 @@ def get_events_generator(descriptor, event_col, descriptor_col,
             if convert_arrays:
                 if _dk['dtype'] == 'array' and not _dk.get('external', False):
                     ev['data'][k] = np.asarray(ev['data'][k])
+
+        # note which keys refer to dereferences (external) data
+        ev['filled'] = filled
+
         # wrap it in our fancy dict
         ev = doc.Document('Event', ev)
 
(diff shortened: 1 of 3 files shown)


tacaswell:
Does this also require a change to the document spec package?

This needs some prose docs.

attn @arkilic I think this should be transparent to the service/client, but just checking

ghost:
itnshould be transparent, I'll try it out. Aside note, we better check how this affects the event retrieval performance

danielballan:
This does require a change to event model if we ever validate on the way out (which we currently never do).

Where do you think these prose docs would go? MDS has none.

I don't expect this to be a significant impact on performance. Since the new dict created once per _descriptor_, not once per event, the cost of merely attaching it to the event should be much less than the cost of writing all the other parts of the event. But sure it doesn't hurt to test.

metadatastore/core.py
```diff
@@ -374,6 +376,10 @@ def get_events_generator(descriptor, event_col, descriptor_col,
             if convert_arrays:
                 if _dk['dtype'] == 'array' and not _dk.get('external', False):
                     ev['data'][k] = np.asarray(ev['data'][k])
+
+        # note which keys refer to dereferences (external) data
+        ev['filled'] = filled
```
tacaswell, line 381: Doesn't this need to be a copy, or the filled state will be implicitly shared across all events.

danielballan, line 381: Yeah, just caught that myself. Fixing.

tacaswell:
it should grow prose docs :smiling_imp: 

At least dump something in https://github.com/NSLS-II/metadatastore/tree/master/doc/api-changes

cowanml:
what is "filled-ness"?
completeness?

danielballan:
https://github.com/NSLS-II/databroker/blob/master/databroker/broker.py#L266

metadatastore/core.py
tacaswell: Did this ever get fixed?  It looks like the same `filled` object is still shared across all of the events.

PR #241, opened 2016-08-09
TST:Simple tests for find_last() and insert()

ghost:


metadatastore/test/test_commands.py
```diff
@@ -504,11 +505,27 @@ def test_double_run_stop():
         mdsc.insert_run_stop(run_start_uid, ttime.time(),
                              uid=str(uuid.uuid4()))
 
+def prep_header():
```
tacaswell, line 511: Can you do this as a fixture?

metadatastore/test/test_commands.py
```diff
@@ -503,12 +504,28 @@ def test_double_run_stop():
...
+    hdr = {'time': time.time(), 'uid': testuid,
+           'tag': 'find_last_test_tag'}
+    return hdr
 
 
 def test_find_last_for_smoke():
     last, = mdsc.find_last()
 
 
+def test_find_last(prep_header):
```
tacaswell: These should also take in the `mds_all` fixture so that these tests run against all three of the mds versions.

ghost: Then shouldn't `test_insert_run_start()` and all other tests also take in the `mds_all` fixture?

tacaswell: Almost all of these are already duplicated in `test_mds` (just not removed here).

PR #242, opened 2016-08-16
API: remove v0 compatibility layer

tacaswell:
If you still need to access an old database (none of which should exist
on the floor) use an old version of mds to migrate to v1.

attn @hhslepicka @danielballan


tacaswell:
@arkilic  is sitting next to me and is happy.

ghost:
`The command "curl -sSL "http://keyserver.ubuntu.com/pks/lookup?op=get&fingerprint=on&search=0x9ECBEC467F0CEB10" | sudo -E apt-key add -" failed and exited with 2 during .`
Travis is being weird, nothing wrong with tests

klauer:
Kicked travis and tests are now passing

PR #244, opened 2016-08-21
Create datasets as empty string 

licode:
Dataset is taken as empty string due to detector defects. This is also used for suitcase test https://github.com/NSLS-II/suitcase/pull/21.


danielballan:
This looks like a good fix to the sample data, but can you help me understand how the current code leads to an empty string?

licode:
This is a bit confusing. I actually want to add data with str type. It looks like in old suitcase we can only save numerical data. That's why we need to do this change to test str type.

licode:
We don't need to worry about data with empty str.

PR #246, opened 2016-08-24
TST:Insert and find return type tests

ghost:


metadatastore/test/test_mds.py
```diff
@@ -159,9 +160,9 @@ def test_insert_run_start(mds_all):
     run_start_uid = mds.insert_run_start(
         time, beamline_id=beamline_id,
         scan_id=scan_id, uid=str(uuid.uuid4()), **custom)
-
+    type(run_start_uid) == str
```
tacaswell: These are lacking an assert.

I would go with `assert isinstance(run_start_uid, six.string_types)`

ghost: One shall never write tests without a cup of coffee

PR #247, opened 2016-08-26
ENH:Authenticate on mongo user. Shells can now be locked

ghost:
Tests need to be updated for this to get merged. Default option is auth=True. This means the mongo instance we are starting in travis and the config used for running tests must have authentication capability enabled.

### DIFF
--- a/metadatastore/mds.py
+++ b/metadatastore/mds.py
@@ -10,7 +10,7 @@ _API_MAP = {1: core}
 
 
 class MDSRO(object):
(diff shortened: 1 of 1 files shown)


tacaswell:
Can you run one of them (maybe the 2.7) with auth off?

ghost:
My goal is to test it against two different mongo daemons one with auth and one w/o auth.

ghost:
Another PR in the works

PR #248, opened 2016-08-30
Authenticate mongoserv

tacaswell:


.travis.yml
```diff
@@ -1,28 +1,31 @@
...
+    - mongodb-org
 
 python:
   - 2.7
   - 3.4
   - 3.5
 
 before_install:
-  - git clone https://github.com/nsls-ii/nsls2-ci ~/scripts
+  - git clone https://github.com/arkilic/nsls2-ci ~/scripts
```
tacaswell: can you make a PR with these changes and then revert this line back?

ghost: See https://github.com/NSLS-II/nsls2-ci/pull/4

tacaswell:
Coverage went down because we removed covered lines of code, but did not create any new misses.

tacaswell:
Can you also add a note in the api_changes folder?

Something like

API: remove mds.commands & mds.api + add auth
- Remove the singleton based API to metadatastore, use MDS / MDSRO instances instead
- require auth when connecting to mongo by default

ghost:
@tacaswell Good to go

tacaswell:
@arkilic Can you please fix the bluesky (and I think ophyd) tests that will be broken by these api changes and update https://nsls-ii.github.io/fresh-installation.html with directions on how to set up authentication on the mongo server?

Why do we need an auth flag?  It would be simpler to just check if a username/password are included in the config?

ghost:
Will do, I just need a day long detox from travis.

The flag is to be able to efficiently toggle between authenticated and open shells. I don't like the idea of just checking for username/pwd since someone can easily forget to populate these fields even when they're required.

ghost:
@tacaswell [This](https://github.com/NSLS-II/bluesky/pull/557) means we don't need to fix the databroker tests?

tacaswell:
Did you link to the wrong thing?  That was my attempt to pre-emptively fix the bluesky tests.

ghost:
Nope, the link is fine. Talked to @danielballan offline yesterday, the confusion is resolved.

PR #250, opened 2016-09-06
TST: Test that exceptions are included in mds object API.

danielballan:
The main point of making this a formal unit test is ensuring that _other_ MDS implementations don't forget to expose the exceptions as attributes.

### DIFF
--- a/metadatastore/test/test_mds.py
+++ b/metadatastore/test/test_mds.py
@@ -417,6 +417,12 @@ def test_fail_runstart(mds_all):
         mdsc.run_start_given_uid('aardvark')
 
 
+def test_exceptions_are_mds_attributes(mds_all):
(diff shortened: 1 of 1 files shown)


ghost:
Not as obsessed about coverage as @tacaswell lol

danielballan:
Haha, no problem, @arkilic. I only added these because I realized I want them for portable-mds. This kind of coverage doesn't mean much for metadatastore itself.

ghost:
I imported them in mdclient like a crazy person as well. mdclient has mds as a requirement just because of these exceptions.

danielballan:
As soon as this is merged I will kick off a PR into portable-mds that depends on this being in MDS master.

ghost:
I will do the same in mdclient

PR #251, opened 2016-09-07
TST: hide explicit MDS imports

tacaswell:
This allows re-use of tests  without requiring pymongo


tacaswell:
This should fix portable mds issues.

PR #253, opened 2016-09-28
API: auth=False by default until it is configurable

danielballan:
There is no way to configure auth is the connection.yml, which means all analysis envs are broken by the default `auth=True`. We had better roll that back to be conservative. In a separate PR we can make auth configurable via the config files, test that it doesn't break anything, and then change the default back to True.

### DIFF
--- a/metadatastore/mds.py
+++ b/metadatastore/mds.py
@@ -10,7 +10,7 @@ _API_MAP = {1: core}
 
 
 class MDSRO(object):
-    def __init__(self, config, version=1, auth=True):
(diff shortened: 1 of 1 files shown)


danielballan:
This isn't even enough. MDS still fails. Stand by for more fixes. ~~Really wish more people came to actual beamlines to actually test things.~~

danielballan:
Sorry, that was unnecessarily snarky. I'm frustrated by how many things are broken at deploy-time.

danielballan:
Will add test without auth on Travis to ensure we don't trip on this next time.

PR #255, opened 2016-11-04
TST: Fixturize the pickle test.

danielballan:
Even if other metadatastore implementations don't support this (and I guess they can...) this should be implement with the fixture.


danielballan:
I did this a very dumb way -- rushing. Will revisit.

Issue #257, opened 2016-11-08
CursorNotFound at MDS


hhslepicka:
At LIX, going over the files of a mesh scan (~4000 files) we received a CursorNotFound exception, result of a timeout at the PyMongo call to find.
Here is the notebook with the complete error message:
http://nbviewer.jupyter.org/gist/hhslepicka/92a84e78534498adf0510a9c4afc5efd

If I do the same but not processing the data all works fine. This is clearly an error caused by a Timeout at the server side of mongo.
```http://api.mongodb.com/python/1.6/api/pymongo/collection.html#pymongo.collection.Collection.find```

My suggestion is to use ```find(..., timeout=False)``` and ensure that the cursor is closed at the end.

Attn. @arkilic @danielballan @tacaswell @lyang11973

**Comment by ghost** [comment]
I've added this a while ago but apparently got lost in one of many rewrites. pymongo claims to be fixing this issue in its latest release but it would be safer to just add it.

**Comment by hhslepicka** [comment]
@arkilic that would be great if you could implement those fixes at all mongo related apps for us while this pymongo fix is not available. Thanks! 👍

**Comment by ghost** [comment]
@hhslepicka will do and you get to enjoy it if we both don't get banned or deported 💨 

```
no_cursor_timeout (optional): if False (the default), any returned cursor is closed by the 
server after 10 minutes of inactivity. 
If set to True, the returned cursor will never time out on the server. 
Care should be taken to ensure that cursors with no_cursor_timeout 
turned on are properly closed.
```

So your request takes more than 10 minutes (I recall @tacaswell and I have decided was enough time and let pymongo cursor manage it automatically). Btw, the docs you are referring belong to pymongo 1.6, you should be getting pymongo 3+ via conda, if not we should fix that.
If we don't close cursors, it might eventually cause memory leakage on the server side. It is a bit more complicated than setting `no_cursor_timeout=True`,  we must close the cursor when the cursor is completely consumed as docs suggest.

PR #258, opened 2016-11-09
ENH:Keep event cursor alive until completely consumed

ghost:
Attn. @tacaswell @hhslepicka 

See issue #257

### DIFF
--- a/metadatastore/core.py
+++ b/metadatastore/core.py
@@ -1029,7 +1029,8 @@ def find_events(start_col, start_cache,
     col = event_col
     events = col.find(kwargs,
                       sort=[('descriptor', pymongo.DESCENDING),
-                            ('time', pymongo.ASCENDING)])
+                            ('time', pymongo.ASCENDING)],
+                      no_cursor_timeout=True)
 
     for ev in events:
         ev.pop('_id', None)
@@ -1043,7 +1044,7 @@ def find_events(start_col, start_cache,
         # wrap it our fancy dict
         ev = doc.Document('Event', ev)
         yield ev
-
+    events.close()
 
 def find_last(start_col, start_cache, num):
     """Locate the last `num` RunStart Documents
--- a/metadatastore/core.py
+++ b/metadatastore/core.py
@@ -1032,19 +1032,21 @@ def find_events(start_col, start_cache,
                             ('time', pymongo.ASCENDING)],
                       no_cursor_timeout=True)
 
-    for ev in events:
-        ev.pop('_id', None)
-        # pop the descriptor oid
-        desc_uid = ev.pop('descriptor')
-        # replace it with the defererenced descriptor
-        ev['descriptor'] = descriptor_given_uid(desc_uid, descriptor_col,
-                                                descriptor_cache,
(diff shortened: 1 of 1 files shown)


metadatastore/core.py
```diff
@@ -1043,7 +1044,7 @@ def find_events(start_col, start_cache,
         # wrap it our fancy dict
         ev = doc.Document('Event', ev)
         yield ev
-
+    events.close()
```
tacaswell: Can you wrap this whole thing in a `try...finally` or is there a context manager for this sort of thing (quick google search says no).

ghost: what if we just do `del events` instead? By instead, I mean closing the cursor part

tacaswell: it is better to be explicit about closing down resources rather than relying on GC and presumably the `__del__` method to take care of it.

ghost: If a cursor is closed and `close()` method is called on this closed cursor, no exception is raised. 

```
n [5]: conn.test.mock.find()
Out[5]: <pymongo.cursor.Cursor at 0x104907748>

In [6]: a = conn.test.mock.find()

In [7]: a.close()

In [8]: a.close()
```

https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/cursor.py#L276

tacaswell:
``` python
try:
     for ev in events:
￼          ev.pop('_id', None)
         # pop the descriptor oid
         desc_uid = ev.pop('descriptor')
         # replace it with the defererenced descriptor
         ev['descriptor'] = descriptor_given_uid(desc_uid, descriptor_col,
                                                 descriptor_cache,
                                                 start_col, start_cache)

￼          # wrap it our fancy dict
￼          ev = doc.Document('Event', ev)
￼          yield ev
finally:
    events.close()
```

tacaswell: Ideally they would have a context manager on find so you could do

``` python
with foo.find(...) as events:
    for ev in events:
        yield ev
```

but I could not find that in the API quickly.

ghost: I really don't think this is necessary given the way that for loop consumes cursors but I will add it since it makes no difference.

ghost: It is not in the API because people are happy with the for loop @tacaswell lol

licode:
safer way to close. LGTM.

tacaswell:
As a side note, the `finally` block also hooks into the `__del__` behavior of generators

``` python
def my_gen():
    try:
        for j in range(5):
            print('in loop', j)
            yield j

    finally:
        print('finally')
```

```
In [3]: list(my_gen())
in loop 0
in loop 1
in loop 2
in loop 3
in loop 4
finally
Out[3]: [0, 1, 2, 3, 4]

In [4]: g = my_gen()

In [5]: next(g)
in loop 0
Out[5]: 0

In [6]: next(g)
in loop 1
Out[6]: 1

In [7]: del g
finally

```

ghost:
If you dig into the __die() method in pymongo driver, it does something similar. It doesn't hurt being safe

ghost:
Closes #257

PR #265, opened 2017-04-02
TST: add test with double start insert

CJ-Wright:


CJ-Wright:
This is more important for the portable-mds stuff, but portable-mds imports all it's tests from mds so...

CJ-Wright:
@tacaswell @danielballan @arkilic Ready for review. (failing on 3.4 for 3.4 reasons)

CJ-Wright:
@tacaswell thank you!

PR #266, opened 2017-05-01
WIP: Listify numpy arrays

ghost:
Not ready for merge. 

In accordance with https://github.com/NSLS-II/bluesky/pull/674

- Removes the deprecated example, not the 'sample_data' (should improve coverage).
- Listifies the numpy arrays on the way in.
- Tests that numpy arrays are inserted within individual and bulk event data fields.
- Tested with bs and ophyd master.

TODO: 
- Needs to be tested with the bluesky and databroker once corresponding prs are merged.
- Update metadataclient for ISS.


tacaswell:
https://github.com/NSLS-II/bluesky/blob/b92e1a42f188fd7efa982c3ac1c8a064c1879a50/bluesky/utils.py#L837 is how we are currently doing this.   This has the advantage of not copying lists.

metadatastore/mds.py
```diff
@@ -656,6 +657,13 @@ def insert_descriptor(self, run_start, data_keys, time, uid, **kwargs):
                                            data_keys=data_keys,
                                            time=time, uid=uid,
                                            **kwargs)
+    def _sanitize_np(self, val):
```
danielballan: Since this does not need access to `self` I'd suggest making it top-level function rather than a method. It's easier to test that way.

tacaswell (approved):
The tests do not test putting in an array, might be worth changing one of the keys to that, but not going to hold the PR over it.

PR #269, opened 2017-07-07
DOC: Finding resources for datum using filestore and metadatastore 

licode:


ghost:
Used and verified at HXN 6/28.

